PR #1, opened 2015-02-01
Add code skeleton and virtualenv

epelz:
Adds skeleton for the code structure, using https://github.com/RuddockHouse/RuddockWebsite
as a base. Additionally, sets up virtualenv and includes instructions for
virtualenv in `README.md`

Completes:
https://app.asana.com/0/24556964288644/25694911605677
https://app.asana.com/0/24556964288644/25190037185565

### DIFF
--- a/Donut/__init__.py
+++ b/Donut/__init__.py
@@ -1,10 +1,10 @@
-from flask import Flask, g
-from sqlalchemy import create_engine
+import flask
+import sqlalchemy
 
 from Donut import config, constants
 from Donut.modules import example
 
-app = Flask(__name__)
+app = flask.Flask(__name__)
 app.debug = False
 app.secret_key = config.SECRET_KEY
 
@@ -17,20 +17,20 @@ app.register_blueprint(example.blueprint, url_prefix='/example')
 # Create database engine object.
 # TODO##DatabaseWork: We currently don't have a database set up, so we can't
 # reference sqlalchemy yet. However, it serves as a good example implementation.
-# engine = create_engine(config.DB_URI, convert_unicode=True)
+# engine = sqlalchemy.create_engine(config.DB_URI, convert_unicode=True)
 
 @app.before_request
 def before_request():
   """Logic executed before request is processed."""
   # TODO#DatabaseWork uncomment this line
-  # g.db = engine.connect()
+  # flask.g.db = engine.connect()
 
 @app.teardown_request
 def teardown_request(exception):
   """Logic executed after every request is finished."""
   # TODO#DatabaseWork uncomment these lines
-  # if g.db != None:
-  #  g.db.close()
+  # if flask.g.db != None:
+  #  flask.g.db.close()
 
 # After initialization, import the routes.
 from Donut import routes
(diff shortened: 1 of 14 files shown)


Donut/templates/layout.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Donut</title>
+    <link rel="stylesheet" type="text/css" 
+          href="{{url_for('static', filename='css/style.css')}}" />
```
allisonkong: Spaces around the `{{` and `}}`?

Donut/__init__.py
```diff
@@ -0,0 +1,36 @@
+from flask import Flask, g
+from sqlalchemy import create_engine
```
allisonkong: Do we want to continue using 

```
from flask import Flask, g, url_for, render_template, ...
from sqlalchemy import text, create_engine, ...
```

Or do we just want to do this:

```
import flask
import sqlalchemy
...

flask.render_template(...)
etc
```

That means we need to prefix things with `flask.`, but we don't need to worry about keeping the import statement updated with whatever functions/variables we're using. At least with flask, we use so many different functions that I think it'll be much easier to just `import flask`. With SQLAlchemy, I'm not sure since we haven't really used ORM before (with the Ruddock website doing `from sqlalchemy import text` is fine since that's literally the only function we ever use).

Donut/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+app.register_blueprint(example.blueprint, url_prefix='/example')
+
+# Create database engine object.
+# TODO##DatabaseWork: We currently don't have a database set up, so we can't
+# reference sqlalchemy yet. However, it serves as a good example implementation.
+# engine = create_engine(config.DB_URI, convert_unicode=True)
+
+@app.before_request
+def before_request():
+  """Logic executed before request is processed."""
```
allisonkong, line 24: :+1: I looked up the style guide for docstrings and this looks good. I think we should stick with these styles:

```
"""This is a short docstring."""
```

or

```
"""
This is a multiple line
docstring that spans
additional lines.
"""
```

The one I think looks really weird is this (let's not use it):

```
"""This is a
multi line docstring.
"""
```

epelz, line 24: Agreed. We'll probably have to change the style somewhat once we add documentation generation, since we'll have to add in more details (e.g. arguments, return). But I'll add the details in once I set that up.

Donut/__init__.py
epelz: Yeah, I'm not really sure. All the flask documentation that I've seen uses `from flask import ...`, which is why we've stuck with it. I kind of like the explicitness of the first approach, though I agree that it becomes a little cumbersome with a lot of functions. Does PEP have a preference?

allisonkong: PEP doesn't really care as long as it's not `from module import *`. It's really up to us, so if `import flask` and `flask.function()` seems cleaner let's go with that.

epelz: Cool, let's try this out then. I'll keep the the blueprint init files (e.g. `Donut/modules/example/__init__.py`) the same, since we'll always only need `flask.Blueprint`.

allisonkong:
looks good to me

PR #7, opened 2015-09-09
Initial

alphaz99:
Travis fixed for meantime without DB. Merging this to provide base for further work (next steps are to fix DB).


Donut/templates/donut.html
```diff
@@ -0,0 +1,24 @@
+{% extends "layout.html" %}
+{% block body %}
+
+<div class="container theme-showcase" role="main">
+
+    <!-- Main jumbotron for a primary marketing message or call to action -->
+    <div class="jumbotron">
+        <img src="static/images/newdonut.png" alt="d(O)nut" title="d(O)nut" style="margin: auto; display: block">
```
allisonkong, line 8: The `url_for('static', filename='images/newdonut.png')` syntax is generally better. For static files it's probably fine to hard code the url, but for other endpoints you definitely want to use the `url_for` function. It helps a lot if you decide to redo how the APIs are laid out.

Donut/templates/layout.html
```diff
@@ -1,31 +1,48 @@
...
-    </div>
-    {% include "includes/footer.html" %}
-  </body>
-  </center>
+    <head>
+        <title>Donut</title>
+        
+        <link rel="stylesheet" type="text/css" 
+              href="{{ url_for('static', filename='css/style.css') }}" />
+        <!-- Google Fonts -->
```
allisonkong, line 8: You can use `{# ... #}` for comments as well. The difference is that HTML style comments are passed on to the user when the page loads, while `{# ... #}` are simply ignored by the template engine so they appear in source code but aren't sent with every request. Minor optimization, but you really don't need to send comments as part of a page's HTML.

Donut/templates/layout.html
```diff
@@ -1,31 +1,48 @@
...
+        <link href='http://fonts.googleapis.com/css?family=EB+Garamond|Roboto' rel='stylesheet' type='text/css'>
```
allisonkong, line 9: https?

Donut/tests/test_routes.py
```diff
@@ -7,4 +7,4 @@ def testHome(client):
   rv = client.get(flask.url_for('home'))
 
   assert rv.status_code == 200
-  assert "Hello world!" in rv.data
+# assert "Hello world!" in rv.data
```
allisonkong, line 10: Maybe it would be better to check that strings like `"<html"` and `"<body"` exist? Not a huge issue, it the status is 200 then it's probably fine. Also code is more readable if you `import httplib` and then use the constants there, like `httplib.OK`, etc.

https://docs.python.org/2/library/httplib.html

run_server.py
```diff
@@ -3,5 +3,5 @@
 # settings enabled.
 from Donut import app, config
 
-port = getattr(config, 'PORT', 5000)
+port = getattr(config, 'PORT', 5050)
```
allisonkong, line 6: If you're having issues with multiple people testing at once, you can individually change the port in the config file (`config.py` isn't part of the codebase so you can change it arbitrarily)

PR #8, opened 2015-09-10
User auth

dqu123:
Basic user authentication / account creation.


Donut/templates/includes/header.html
```diff
@@ -61,7 +61,7 @@
                         </ul>
                     </li>
                     <li>
-                        <form action="#">
+                        <form action="/login">
```
allisonkong: `<form action="{{ url_for('auth.login') }}">`

PR #11, opened 2015-09-23
User auth merge

dqu123:
Merged user auth, added arg parser to run_server.py (to specify port).
@dkong1796, @alphaz99, @ootks, @Timeroot, @ChingYunH, @RobertEng, could
you take a look so we can merge? 

One question I had was why did you move creation
of the engine variable in Donut/**init**.py  inside of the before_request function, Rahul?
(It should be more efficient to only make the engine once instead of for each request).


run_server.py
```diff
@@ -2,9 +2,17 @@
 # uses wsgi to start up, and bypasses this file. So we are free to have debug
 # settings enabled.
 from Donut import app, config
+import argparse
+
+parser = argparse.ArgumentParser(description='Set up development server.')
+parser.add_argument('port_num', metavar='port_num', type=int,
```
allisonkong: Why not call it 'port'?

run_server.py
```diff
@@ -2,9 +2,17 @@
...
+parser = argparse.ArgumentParser(description='Set up development server.')
+parser.add_argument('port_num', metavar='port_num', type=int, 
+                    help='Port on which to run server.')
+
+args = parser.parse_args()
+
 
 if __name__ == "__main__":
-    port = getattr(config, 'PORT', 5050)
+    port = getattr(config, 'PORT', args.port_num)
```
allisonkong: Presumably if you pass in a port, you really want to use that port. `getattr(config, 'PORT', args.port_num)` will always use `config.PORT` unless that constant is not set, in which case it uses the one passed in by a flag, so it will basically always ignore the port provided.

allisonkong:
@dqu123  are you referring to this?
https://github.com/RuddockHouse/RuddockWebsite/pull/87

I did that because we kept on getting "MySQL server has gone away" exceptions if the server was running but no one had made a request for a few hours (default is 8 hours). Theoretically sqlalchemy's pool refresh should have taken care of the issue, but we tried it and were still getting errors so I just did this as a quick hack to fix it (I'm definitely not suggesting that this is the best solution).

alphaz99:
That makes sense - I wasn't fully sure why it was written like that either but until we figure out a better solution we can keep this as a temporary fix.

run_server.py
allisonkong: Also you can configure argparse so that `python run_server.py -p 5000` and `python run_server.py --port 5000` do the same thing, following unix convention.

dqu123:
Fixed the port argument. Yeah I kinda committed it like instantly without analyzing what getattr was doing; thanks for the heads up.

run_server.py
```diff
@@ -2,9 +2,20 @@
...
+parser = argparse.ArgumentParser(description='Set up development server.')
+parser.add_argument('-p', '--port', metavar='port', type=int, 
+                    help='Port on which to run server.')
+
+args = parser.parse_args()
+
 
 if __name__ == "__main__":
-    port = getattr(config, 'PORT', 5050)
+    if args.port:
```
allisonkong: Testing if a value is not none should always be done with `if args.port is not None`, since that makes it unambiguous. If I passed in `--port 0` as the arg, this would evaluate to false incorrectly (not that 0 is a valid port, but you get the idea).

dqu123:
Thanks for the nit-picking @dkong1796. We definitely need to keep up the code reviews so we can have a high standard of code quality.

allisonkong:
No problem :p

Everything else looks fine (minus the parts that aren't implemented haha). I haven't actually tested anything, only looked at the code.

allisonkong:
Actual misc question: do we want to keep on using `Donut` as the name of the main package, or use `donut`? The Python PEP8 standard prefers all lowercase:
https://www.python.org/dev/peps/pep-0008/#package-and-module-names

alphaz99:
donut lowercase sounds good to me

dqu123:
Renamed. We should go ahead and merge if it looks good, so master has the corrected module name.
btw I used the command `find . -type f -name "*.py" -exec sed -i 's/Donut/donut/g' {} +`

allisonkong:
Ideally all the python3 stuff should have taken place in a different branch/pull request (it's not related to the topic of this pull request). Are we definitely switching to python3 then? I was mostly suggesting it as something to be explored in #12 so we could identify issues before committing to switching.

dqu123:
Moved Python 3 stuff to #13 and reverted the stuff on this branch.
EDIT: Whoops messed up the revert, I'll go fix that.

.gitignore
```diff
@@ -1,4 +1,4 @@
-Donut/config.py
+donut/config.py
```
allisonkong, line 1: Looks like the rename was reverted in this branch. If you just have `config.py` in the `.gitignore` then it'll ignore any file that is named `config.py`, regardless of which directory it is in.

allisonkong:
:+1:

dqu123:
Yeah wait I think I accidentally did the wrong thing when I reverted, because revert lets you revert a specific commit, but I intended to revert back up to that commit.

allisonkong:
Oh, haha. `git reset --hard <commit hash>` does what you want, but then you'd have to `git push --force` to update the upstream repo. You would also lose all the commits in between. I'm not familiar enough with git to suggest something more elegant (you _could_ revert each commit manually, and then revert the reverts to restore them...)

Edit: Actually, `git checkout -b` on a new branch with all the commits you do want (so you don't lose them) followed by a `git reset --hard` on the old branch might do the trick.

dqu123:
Yeah I accidentally did `git revert <hash>` which just reverted the hash. But this is the best way to learn git right... lol

dqu123:
Is there a good way to reset it on the Github side? Or do I have to delete the branch and repush it?

allisonkong:
Does `git push origin user_auth --force` not work?

dqu123:
Okay, we're good now. I'll keep the `--force` bit in mind next time.

allisonkong:
:D

donut/tests/test_routes.py
```diff
@@ -1,7 +1,7 @@
 import flask
 
-from Donut.tests.fixtures import client
-from Donut import app
+from donut.tests.fixtures import client
+from donut import app
 
 def testHome(client):
```
allisonkong: This should be renamed `test_home` to follow convention.

dqu123:
Hmm, it would be nice if we could version control the config.py since it might be slightly different depending on the mysql connector.

dqu123:
I think we can go ahead and merge this. This pull is a bit scattered since it includes a lot of fundamental changes, but I just did it in one branch since there were no other active ones at the time (I plan on making an upload script branch and another branch to document my progress on the schema). I'll also open up another pull request for user_auth that will focus more specifically on implementing user authentication.

I'm definitely seeing distinct roles that can be done in parallel now, so I hope to set that in motion and assign tasks during our first meeting. For example, working on the front-end for the directory search,  writing the backend user_auth logic, writing the registrar upload script, and writing SQL DDL. I'll try to get the basic SQL needed for the backend logic and script to be started

PR #13, opened 2015-09-26
Py3

dqu123:
Branch for potential Python 3 move.


dqu123:
@dkong1796  @alphaz99 Hmm, it would be nice if we could version control the config.py since it might be slightly different depending on the mysql connector.

allisonkong:
Yeah... I really don't want to upload `config.py` anywhere (even in a private repo) since that introduces a huge security risk. Maybe we could do something with command line flags to dynamically choose which db connector to use? Not really sure about this one.

dqu123:
I guess we can just name the python3 one a different name. It would be nice to just have it in 
version control, since I'm so used to tracking my changes that way, but I understand the security vulnerability. (Also version control would help avoid deleting it when you rename the directory)

epelz:
One option is that `config.py` can read environment variables which store the secure things, so that you can commit the config file without exposing the secrets. Then you'd just have the server export the secure variables

alphaz99:
I actually kind of like that idea a lot. Sounds simple and elegant, and that makes sense too.

dqu123:
Ultimately we will still need some secret files not in version control that will break `run_server.py` if 
you don't have them so I think it's equivalent to just have hidden config files with different names.

epelz:
Not quite - the config file then serves as a template of which secrets you need and you can version control as you change its format (which is what you said you wanted).

dqu123:
See #35

PR #14, opened 2015-09-27
[WIP] Sql

dqu123:
Using this pull request to track SQL issues.

### DIFF
--- a/sql/donut.sql
+++ b/sql/donut.sql
@@ -1,0 +1,48 @@
+DROP TABLE IF EXISTS users;
+DROP TABLE IF EXISTS members;
+
+CREATE TABLE members (
+    user_id int(11) NOT NULL AUTO_INCREMENT, -- Primary Key
+    uid char(7) NOT NULL,  -- Unique
+    last_name varchar(255) NOT NULL, 
+    first_name varchar(255) NOT NULL,
+    middle_name varchar(255), 
+    email varchar(255) NOT NULL,
+    phone varchar(64) DEFAULT NULL,
+    gender tinyint(2) DEFAULT NULL, -- Numerical code that will be stored in code
+    birthday date DEFAULT NULL,
+    entry_year year(4) DEFAULT NULL,
+    graduation_year year(4) DEFAULT NULL, -- Question: does this 1-to-1 denote year of school (frosh, soph)
+    msc smallint(6) DEFAULT NULL,
+    building varchar(32) DEFAULT NULL,
+    room_num smallint(6) DEFAULT NULL,
+    is_abroad bool DEFAULT 0,
+    major varchar(16) DEFAULT NULL,
+    major2 varchar(16) DEFAULT NULL,
+    address varchar(255) DEFAULT NULL, -- Someone confirm how big this varchar should be.
+    city varchar(64) DEFAULT NULL,
+    state varchar(64) DEFAULT NULL,
+    zip varchar(9) DEFAULT NULL,
+    country varchar(64) DEFAULT NULL, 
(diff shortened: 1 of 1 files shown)


dqu123:
I'll figure out how to get travis to ignore this branch later.
Currently, member table is in donut database. But user table failed to work because of the NOW() issue.

sql/donut.sql
```diff
@@ -0,0 +1,42 @@
...
+    major varchar(16) DEFAULT NULL,
+    major2 varchar(16) DEFAULT NULL,
+    address varchar(255) DEFAULT NULL, -- Someone confirm how big this varchar should be.
+    city varchar(64) DEFAULT NULL,
+    state varchar(64) DEFAULT NULL,
+    zip varchar(9) DEFAULT NULL,
+    country varchar(64) DEFAULT NULL, 
+    create_account_key char(32) DEFAULT NULL,
+    PRIMARY KEY (member_id),
+    UNIQUE KEY uid (uid)
```
allisonkong: I don't think you need to name the constraints. This creates a constraint named "uid" - if you just did `UNIQUE (uid)` then MariaDB would just automatically generate a name for the constraint.

sql/donut.sql
```diff
@@ -0,0 +1,42 @@
+DROP TABLE IF EXISTS members;
+CREATE TABLE members (
+    member_id int(11) NOT NULL AUTO_INCREMENT, -- Primary Key
+    uid char(7) NOT NULL,  -- Unique
+    last_name varchar(255) NOT NULL, 
+    first_name varchar(255) NOT NULL,
+    middle_name varchar(255), 
+    email varchar(255) NOT NULL,
+    phone varchar(64) DEFAULT NULL,
+    gender enum('m', 'f') DEFAULT NULL, -- Question: are there PC options we need to add?
```
allisonkong: This should be an integer. The enum should be stored as part of the codebase, not part of the database, so something like this:

```
class Gender(object):
  UNKNOWN = 0
  MALE = 1
  FEMALE = 2
  OTHER = 3
```

Which you can then use directly in the code (it's more clean than directly comparing strings). 

And you should separately have another column that optionally stores a string for when their gender is `Gender.OTHER`. Pronouns would be they, their, etc.

sql/donut.sql
```diff
@@ -0,0 +1,42 @@
...
+    city varchar(64) DEFAULT NULL,
+    state varchar(64) DEFAULT NULL,
+    zip varchar(9) DEFAULT NULL,
+    country varchar(64) DEFAULT NULL, 
+    create_account_key char(32) DEFAULT NULL,
+    PRIMARY KEY (member_id),
+    UNIQUE KEY uid (uid)
+);
+
+DROP TABLE IF EXISTS users;
```
allisonkong: If you tried to run the script twice in succession right now, it would fail because `users` has a FK to `members`, so your `DROP TABLE IF EXISTS members` statement would fail (well, it might not fail if the users table is empty, but if there _were_ rows then it would fail).

allisonkong:
Relevant: https://github.com/RuddockHouse/RuddockWebsite/pull/110

allisonkong:
Looks good so far! Can we keep everything contained to one place, though (I noticed there's some stuff in the wiki too)?

dqu123:
Yeah I removed the corresponding tables in the wiki. I'll try to just keep things in this branch for simplicity. But the stuff in the wiki right now is not in the branch. We can move it all to this branch to keep it organized.

dqu123:
Hmm actually, I'll probably bring back the travis if I can find a SQL test framework. Does anyone know one? @dkong1796 @epelz @alphaz99

PR #16, opened 2015-09-28
Registrar Upload script

dqu123:
<h3>Registrar Upload script specifications</h3>

```upload_registrar.py ascit_2015.txt```

- [ ] Be able to work on the same file twice
- [x] Insert new members (UID doesn't exist in db)
- [ ] Update old members (UID does exist in db)
- [ ] Handle case of person without UID (Probably should report this to stdout.)
    And other null cases. (See schema to check which columns are NOT NULL).
- [ ] Make the script easy to modify (since the schema is not set in stone)
- [ ] Give option to load to test db
- [ ] Write test script for upload script
- [x] Initially test on sample data (Feel free to add fake test cases)
- [ ] Test on Fall 2015 data. (DO NOT COMMIT actual data to git)

### CODE REVIEW

**Comment by dqu123** [comment]
@RobertEng Yeah maybe we should remake this pull request to take out all the extraneous commits.

**Comment by dkong1796** [comment]
You don't need to remake the pull request - you just needed to merge master :-)

**Comment by RobertEng** [comment]
Can this be merged @dqu123?

**Comment by dqu123** [comment]
@RobertEng, this still needs a lot of work, but you can take a look at https://github.com/ASCIT/donut-legacy/blob/master/utils/directory_update_script.py and its dependencies if you want to work on it (I haven't used this script, but @alphaz99 has so he might have more insights on it). I'll be implementing user authentication first, but will try to come back to this and look at it again.

File: upload/upload_registrar.py [hunk 1]
```diff
@@ -1,8 +1,141 @@
...
+    room_num = Column(Integer)
+    address = Column(String)
+    city = Column(String)
+    state = Column(String)
+    zip = Column(String)
+    country = Column(String)
+    create_account_key = Column(String)
+    
+    def __repr__(self):
+        return "<User(uid=\"%s\", last_name=\"%s\", first_name=\"%s\")" %\
```
dqu123: Using `"{}".format(some_value)` is cleaner than `"%s" % some_value`
Note you can number them: 
`"{0} {1} {0}".format("hello", "world") == "hello world hello"` 
There's a few places you use %, so make sure to change them all.

upload/upload_registrar.py
```diff
@@ -1,8 +1,141 @@
...
+    if verbose:
+        print "Uploading students."
+    for student in student_data:
+        session.add(student)
+    try:
+        session.commit()
+    except IntegrityError as e:
+        print e
+        print "Rolling back all server changes..."
+        session.rollback()
```
dqu123: Maybe it would be good to prompt "Done uploading. Commit changes? y/n"
and rollback if you select no. Then you could run the script and test the results and rollback if something bad happened.

upload/upload_registrar.py
```diff
@@ -1,8 +1,141 @@
...
+        for raw_line in registrar_file:
+            # use regex to get all entries from between quotes
+            parsed_line = [None if entry == "" else entry
+                for entry in re.findall(r"\"(.*?)\"", raw_line)]
+            # instanciate a new member
+            new_member = Member(
+                last_name = parsed_line[1],
+                first_name = parsed_line[2],
+                middle_name = parsed_line[3],
+                uid = int(parsed_line[4]),
```
dqu123: `uid` should be left as a string.

upload/upload_registrar.py
```diff
@@ -1,8 +1,141 @@
...
+                gender = 0 if parsed_line[5] == "M" else 1,
+                birthday = datetime.datetime
```
dqu123: I'm getting an error on parsing birthday.
It's because there's a non-standard format for months ("SEP").
See this legacy [script](https://github.com/ASCIT/donut-legacy/blob/master/utils/users/update_directory_data.py) for one way to approach it.

upload/upload_registrar.py
```diff
@@ -1,8 +1,141 @@
...
+Args:
+    filename: registrar provided file of new students
+    database: select "dev" or "prod" database
+    -v: increase output verbosity
+
+Written by Michael Huynh
+July 05, 2017
+"""
+
+# import all that good stuff
```
dqu123: This comment isn't necessary

upload/upload_registrar.py
```diff
@@ -76,61 +40,93 @@ def parse_file(filename, verbose = False):
...
-        session.commit()
-    except IntegrityError as e:
-        print e
+        print "Connected."
+        print "Staging changes..."
+        
+    transaction = db_conn.db.begin()
+        
+    for member in member_data:
+        cmd = ("INSERT INTO members (last_name, first_name, middle_name, uid, "
```
dqu123: You should parameterize this query. See https://github.com/ASCIT/donut/blob/master/donut/modules/auth/helpers.py#L24 for an example

PR #19, opened 2015-10-02
Scheduler

dqu123:
Making a PR so we can review the code.


dqu123:
Moved to #33 
@majesticfish leaving this branch in case I forgot some files when I transferred, but we should
delete it when we're sure we got everything and the new branch is working

PR #20, opened 2015-12-16
Donut Schema v0.1

dqu123:
This PR contains the SQL file that represents our new schema.

### DIFF
--- a/sql/donut.sql
+++ b/sql/donut.sql
@@ -68,7 +68,3 @@ CREATE TABLE users (
     UNIQUE KEY (username)
 );
 
-
-/* Test data / initial data */
-INSERT INTO members(uid, last_name, first_name, email)
-    VALUES ('1957540', 'Qu', 'David', 'davidqu12345@gmail.com');
--- a/sql/donut.sql
+++ b/sql/donut.sql
@@ -1,0 +1,48 @@
+DROP TABLE IF EXISTS users;
+DROP TABLE IF EXISTS members;
+
+CREATE TABLE members (
+    user_id int(11) NOT NULL AUTO_INCREMENT, -- Primary Key
+    uid char(7) NOT NULL,  -- Unique
+    last_name varchar(255) NOT NULL, 
+    first_name varchar(255) NOT NULL,
+    middle_name varchar(255), 
+    email varchar(255) NOT NULL,
+    phone varchar(64) DEFAULT NULL,
+    gender tinyint(2) DEFAULT NULL, -- Numerical code that will be stored in code
+    birthday date DEFAULT NULL,
+    entry_year year(4) DEFAULT NULL,
+    graduation_year year(4) DEFAULT NULL, -- Question: does this 1-to-1 denote year of school (frosh, soph)
+    msc smallint(6) DEFAULT NULL,
+    building varchar(32) DEFAULT NULL,
+    room_num smallint(6) DEFAULT NULL,
+    is_abroad bool DEFAULT 0,
+    major varchar(16) DEFAULT NULL,
+    major2 varchar(16) DEFAULT NULL,
+    address varchar(255) DEFAULT NULL, -- Someone confirm how big this varchar should be.
+    city varchar(64) DEFAULT NULL,
+    state varchar(64) DEFAULT NULL,
+    zip varchar(9) DEFAULT NULL,
+    country varchar(64) DEFAULT NULL, 
+    create_account_key char(32) DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


sql/donut.sql
```diff
@@ -0,0 +1,48 @@
+DROP TABLE IF EXISTS users;
+DROP TABLE IF EXISTS members;
+
+CREATE TABLE members (
+    user_id int(11) NOT NULL AUTO_INCREMENT, -- Primary Key
```
allisonkong: The number in parentheses isn't really necessary. It's a hint to the number of base-10 digits that is the longest number that can be stored, and doesn't have any special meaning beyond that.

sql/donut.sql
```diff
@@ -0,0 +1,48 @@
...
+    uid char(7) NOT NULL,  -- Unique
+    last_name varchar(255) NOT NULL, 
+    first_name varchar(255) NOT NULL,
+    middle_name varchar(255), 
+    email varchar(255) NOT NULL,
+    phone varchar(64) DEFAULT NULL,
+    gender tinyint(2) DEFAULT NULL, -- Numerical code that will be stored in code
```
allisonkong: Gender is _not_ binary. You should support a custom gender field and have a separate nullable varchar column that gives a custom gender field in addition to this column.

sql/donut.sql
```diff
@@ -0,0 +1,48 @@
...
+    birthday date DEFAULT NULL,
+    entry_year year(4) DEFAULT NULL,
+    graduation_year year(4) DEFAULT NULL, -- Question: does this 1-to-1 denote year of school (frosh, soph)
```
allisonkong: I don't think there's a good way of determining who is a frosh/sophomore/junior/senior if someone is not on a 4-year graduation schedule. I think this is fine, since we typically do things (e.g. mailing lists) by year (ug-2015, ug-2016, etc) instead of by label (there's no ug-frosh, ug-sophomores, etc).

sql/donut.sql
```diff
@@ -0,0 +1,48 @@
...
+    msc smallint(6) DEFAULT NULL,
+    building varchar(32) DEFAULT NULL,
+    room_num smallint(6) DEFAULT NULL,
+    is_abroad bool DEFAULT 0,
```
allisonkong: ???

I realize this was copied from the ruddock website, but still. Also I'm getting rid of that field there too:
https://github.com/RuddockHouse/RuddockWebsite/pull/110

sql/donut.sql
```diff
@@ -0,0 +1,48 @@
...
+    major varchar(16) DEFAULT NULL,
+    major2 varchar(16) DEFAULT NULL,
```
allisonkong: No. We should have a separate table of possible majors, and then a many-to-many relationship between undergrads and majors (this could also potentially contain minors?)

sql/donut.sql
dqu123: I think it only determines the number that is viewed in command line, but yeah 
I think I copied it a while ago from the Rudd dump, but I'll take it out. It is certainly misleading 
since the size of the integer types is fixed.

sql/donut.sql
dqu123: I think TINYINT is a byte on mariadb. See https://mariadb.com/kb/en/mariadb/tinyint/

sql/donut.sql
dqu123: @dkong1796 what's the best way to populate the table of majors?

allisonkong: Probably just manually go through a catalog and add the majors to your sql script? And then query the old db to see what majors alums were that might have been changed since.

allisonkong: There aren't so many majors that doing anything automated would be faster than a one-time manual thing.

sql/donut.sql
allisonkong: Yeah, this int size is fine. I just meant to also add a string column for gender.

sql/donut.sql
allisonkong: Yeah. Honestly I'd prefer it if the numbers in parentheses were the number of bytes :-\

sql/donut.sql
epelz: +1 to dk

sql/donut.sql
epelz: yeah we currently populate majors by running the "populate student registry" script, and if it crashes because of an undefined major then we add it manually. This is pretty infrequent so it's not a huge burden, but it would be slightly better if the script said "Do you want to add X?".

PR #21, opened 2016-03-23
WIP: Setting up database configuration

dqu123:


dqu123:
How does this look @dkong1796, @RobertEng, @alphaz99 ?
It sets the default config to the public database if there is no config, but 
otherwise uses the development database.

As a reminder, we plan on having a production, development, and test database.
The development will periodically copy from the production while the test database will have
a static dataset for writing unit tests.

I'm working on setting up the user / passwords for the other databases and writing a config.py
file that everyone can use.

allisonkong:
LGTM :P 

You may also want to change the flask client used in testing to also be configured to use the test environment and add the TESTING environment parameter (see http://flask.pocoo.org/docs/0.10/config/#builtin-configuration-values).

allisonkong:
See https://github.com/RuddockHouse/RuddockWebsite/blob/master/ruddock/testing/fixtures.py#L10
and https://github.com/RuddockHouse/RuddockWebsite/blob/master/ruddock/default_config.py#L12

run_server.py
```diff
@@ -1,21 +1,19 @@
...
 import argparse
 
-parser = argparse.ArgumentParser(description='Set up development server.')
-parser.add_argument('-p', '--port', metavar='port', type=int, 
-                    help='Port on which to run server.')
+import donut
+from donut import app
 
-args = parser.parse_args()
+parser = argparse.ArgumentParser(description="Set up development server.")
```
allisonkong: s/Set up/Run/

PR #27, opened 2017-02-14
WIP:  Add marketplace.sql

mvgiambrone:
This is a work in progress, currently trying to duplicate the old marketplace tables in the new database.


allisonkong:
Let me know if you have any questions about marketplace (I'm the original author of the php implementation) or if you want me to review anything :)

tjleing:
> Let me know if you have any questions about marketplace (I'm the original author of the php implementation) or if you want me to review anything :)

Awesome; thank you! :)

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,274 @@
...
+            if edition == 3:
+                return "3rd"
+            return str(edition) + "th"
+        else:
+            return str(edition)
+    except ValueError:
+        return edition
+
+
+def get_name_from_user_id(user_id):
```
dqu123: You can use the `members_full_name` view for this.

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,274 @@
...
+    for column in result:
+        categories.append(column[0])
+
+    # Get number of rows and columns to display categories nicely. This is a
+    # little tricky - we want to aim for a table with either 4 categories or 5
+    # categories per row, with the last row catching the remainder. However, we
+    # want to avoid having exactly 1 category in the remainder row if possible.
+    num_cats = len(categories);
+    num_cols = num_cats
+    if num_cats <= 5:
```
dqu123: This logic is a bit hard to understand. Is it completely necessary or is there a different way of displaying? I'm also not completely sure what the issue is. Maybe a drawing would help.

dqu123 (changes requested):

donut/modules/marketplace/helpers.py
tjleing: Fixed.

dqu123:
@bluezebragames, is there currently a way to remove items if they've been bought?

tjleing:
@dqu123 not yet, and moreover, editing is not supported either (although most of the infrastructure for it is up already).  The reasoning was that both of those features are gated behind logging in as either the creator of the item or as admin, and AFAIK privileges aren't working yet.  Also, since I don't have a donut account, testing that would be difficult.  I could make those features accessible to everyone, with the gating being easily addable later; is that preferable?

tjleing:
### Summary
- These changes implement the marketplace functionality of donut.
- First, users can look at items and can filter listings by category and/or search queries.
	- The search bar supports filtering by words in the title, textbook author, and ISBN.  Just searching with any of those components will return matches.  Perfect matches are displayed, if any, and if not, imperfect matches are.  Certain words, like 'a,' 'all,' and 'am,' are ignored.
- Also, users can add listings to the database, selecting item condition, price, title, and additional details, and if the item is a textbook, ISBN and author.
- The user can pick the textbook that they are selling from a list, which specifies the title and author automatically, and if the textbook they are selling is not available, they can add it.

### Test Plan
- [ ] Make sure that searching works, by selecting individual categories, and querying by words and ISBNs.
- [x] Make sure that queries that return no results are handled properly (and not just an empty table)
- [x] Make sure that items of category textbook and non-textbook can be added successfully.
- [x] Make sure that new textbooks can be added to the database by using the form.
- [ ] Make sure that just by switching pages (using the next and previous buttons) does not erase any data
	- [ ] even if the data has spaces in it, because that was causing trouble earlier
- [x] Make sure that you cannot continue to the next page of the form if any required data is missing, or if any other data is malformed
- [ ] Make sure that even by going into the HTML and changing the hidden form elements, malformed data cannot be inserted into the database
- [ ] Check for SQL injections by inputting apostrophes into search boxes and form boxes and seeing if the app crashes

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
...
+import sqlalchemy
+import re
+
+from donut.modules.core.helpers import get_member_data
+
+import routes
+
+# taken from donut-legacy, which was apparently taken from a CS11
+# C++ assignment by dkong
+skip_words = ["a", "all", "am", "an", "and", "are", "as", "at",
```
dqu123: Should be 
```python 
SKIP_WORDS = ...
```
see https://www.python.org/dev/peps/pep-0008/#constants

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
...
+        return len([x for x in l1 if x in l2])
+    else:
+        return len([x for x in l2 if x in l1])
+
+
+def tokenize_query(query):
+    """
+    Turns a string with a query into a list of tokens that represent the query.
+    """
+    global skip_words
```
dqu123: Is the `global` keyword needed?

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
...
+
+    try:
+        edition = int(edition)
+        if edition < 1000:
+            # it's probably an edition, not a year
+
+            # if the tens digit is 1, it's always "th"
+            if (edition / 10) % 10 == 1:
+                return str(edition) + "th"
+            if edition%10 == 1:
```
dqu123: This is a bit hard to read. Try:
```python
if edition % 10 == 1:
```

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
...
+        fields = all_returnable_fields
+    else:
+        if any(f not in all_returnable_fields for f in fields):
+            return "Invalid field"
+
+
+    # add spaces around method to make the query work
+    method = " " + method + " "
+    # Build the SELECT and FROM clauses
+    s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text(method.join(tables)))
```
dqu123: What happens if you run this with something malicious for `method`?

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
...
+
+    Arguments:
+        datalist: a 2d list of data, with columns determined by fields
+        fields: the column titles from the SQL tables
+
+    Returns:
+        datalist: the original table, but with the two columns merged
+        fields: the column titles similarly merged together into item_title
+    """
+    item_index = -1
```
dqu123: `None` is more readable. You can check 
```python
if index is None:
```

donut/modules/marketplace/routes.py
```diff
@@ -0,0 +1,342 @@
...
+    # 2:  Input information about the listing
+    # 3:  Confirm that info is correct
+    # 4:  Add data to database, show success message
+
+    page = 1 # default is first page; category select page
+    if "page" in flask.request.form:
+        # but if we pass it in, get it
+        page = int(flask.request.form["page"])
+
+    if not page in [1, 10, 2, 3, 4]:
```
dqu123, line 122: Using an enum and adding it to a `constants.py` might be easier.
However, we might want to wait for us to move to Python 3 since that introduces enums natively. 
I'm going to try to get a Python 3 PR merged before this / prio someone get it done
before we merge this PR.

donut/modules/marketplace/routes.py
```diff
@@ -0,0 +1,342 @@
...
+        return helpers.render_with_top_marketplace_bar('view_item.html', stored=stored, cat_title=cat_title)
+
+    except ValueError:
+        return flask.render_template('404.html')
+
+
+@blueprint.route('/marketplace/sell', methods=['GET', 'POST'])
+def sell():
+    # if the data or the page in general has errors, we don't let the user continue to the next page
+    hasErrors = False
```
dqu123: Should be `has_errors`.

donut/modules/marketplace/routes.py
```diff
@@ -0,0 +1,342 @@
...
+        data = helpers.get_table_list_data(['marketplace_items', 'marketplace_textbooks'], storedFields, {'item_id': item_id}, "NATURAL LEFT JOIN")[0]
+        for i in range(len(data)):
+            stored[storedFields[i]] = data[i]
+
+        # cat_title from cat_id
+        cat_id_map = {sublist[0]: sublist[1] for sublist in helpers.get_table_list_data("marketplace_categories", ["cat_id", "cat_title"])}
+        cat_title=cat_id_map[stored["cat_id"]]
+
+        # full_name and email from user_id
+        data = helpers.get_table_list_data(['members', 'members_full_name'], ['full_name', 'email'], {'user_id': stored['user_id']}, "NATURAL LEFT JOIN")[0]
```
dqu123: Since you use the `user_id` from stored, which comes from the first result,
this always uses the same `user_id` (not the `user_id` of the user who is logged in).
You should check if there's a helper function to do this in the core module and add it if there's not.

dqu123 (changes requested):

donut/modules/marketplace/helpers.py
```diff
@@ -0,0 +1,754 @@
+import flask
+import sqlalchemy
+import re
+
+from donut.modules.core.helpers import get_member_data
+
+import routes
+
+# taken from donut-legacy, which was apparently taken from a CS11
+# C++ assignment by dkong
```
allisonkong: Yeah, I got it from here:
http://users.cms.caltech.edu/~donnie/cs11/cpp/lab7/index.html

It looks like Donnie's completely redone the C++ track since

PR #29, opened 2017-03-29
Finished the proposed SQL changes

RobertEng:
In this implementation:
* Houses are in the organization table and NOT a separate table. Instead they are pulled out into a view along with the house members.
* Organizations and groups are separate tables and are in no way connected.

Future:
- Actually test the SQL in the test db to make sure it works. The main purpose of this pull request is to just get the tables down for approval.
- TODO: Add CASCADEs
- TODO: Better comments


majesticfish (approved):
LGTM

PR #34, opened 2017-08-05
Quick fix to REST API

tjleing:


donut/modules/core/helpers.py
```diff
@@ -28,6 +28,9 @@ def get_member_data(user_id, fields=None):
     # Build the SELECT and FROM clauses
     s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("members"))
 
+    # Build the WHERE clause
+    s = s.where(sqlalchemy.text("user_id = " + str(user_id)))
```
dqu123: You should parameterize this query. See https://github.com/ASCIT/donut/blob/master/donut/modules/auth/helpers.py#L24 for an example
I would do `s = s.where(sqlalchemy.text("user_id = :u"))` and `result = flask.g.db.execute(s, u=user_id).first()`. Otherwise the `user_id=user_id` part isn't used. I think this makes it clear what parameters are being passed to be sanitized by putting them all in the same function call instead of scattered through various lines.


We should probably agree on a consistent way to write these queries and write a style guide since
sqlalchemy gives a lot of different options. I think just using `sqlalchemy.text` for static queries (so you can read the sql easily) would be a good policy and only using the `select`, `select_from`, and `where` functions
when you have a dynamic set of fields.

dqu123 (changes requested):

donut/modules/core/helpers.py
allisonkong: +1, good catch. String concatenation like this is the easiest way you can introduce SQL injection vulnerabilities into your system.

dqu123:
lgtm

dqu123 (approved):

PR #39, opened 2017-10-02
Marketplace

tjleing:
I accidentally merged the old marketplace branch into master instead of the other way around, so I'm opening a new pr for the rest of the changes.

### Summary
- These changes implement the marketplace functionality of donut.
- First, users can look at items and can filter listings by category and/or search queries.
- Also, users can add listings to the database, selecting item condition, price, title, and additional details.  If the item is a textbook, ISBN and author can also be selected.
- The user can pick the textbook that they are selling from a list, which specifies the title and author automatically, and if the textbook they are selling is not available, they can add it.
- Once a user creates a listing, they can also go in and edit it, changing any information they want to.

### Remaining to do
- [x] Enums for page names
- [x] I don't actually know if editing works, so that's kinda important
- [x] Let users edit via the item view page
- [x] Fail gracefully if item_id is invalid
- [x] Add tests
- [x] Using the logged-in user's id when submitting items
- [x] Make sure that when a user is editing, the user_id matches
- [x] Only display edit button if the user created the listing or is admin
- [x] Page for users to manage (edit or remove) their items
- [x] Also images exist (or should)

### Various smaller todos:
- [ ] get_table_list_data into _one and _all (or an argument)?
- [x] confirmation page (html and python)
- [ ] various refactoring
- [x] pass images to sell_2
- [x] pass images to sell_3
- [x] actually set things in the sql
- [x] pass images to view
- [x] handle adding and editing images
- [x] flash on inactive item
- [x] range(____|length) in Jinja to loop.index0
- [x] use directory search instead of the random endpoint things
- [ ] figure out table permissions
- [ ] even more tests?
- [x] cursor.lastrowid???
- [x] flexbox for categories instead of jank table
- [ ] use cat_order to sort categories (in top bar, item creation dropdown, etc) to ensure that selecting a category actually goes to the right place
- [ ] make sure that no change needs to be made due to permissions being different now

### CSS stuff to do after that's up (very exciting!)
- [x] tables probably shouldn't go all the way to the edges of the screen -- like 10% padding on each side?
  - [ ] have another look at that when CSS is up
- [ ] what's going on with the footer bar (esp. when creating a new item to sell)
- [ ] stylesheets instead of embedding CSS into HTML objects 😢

### Test Plan
- [x] Use the links to individual categories in the top box and make sure that items show up correctly.
- [x] Search for a few queries to see that querying works correctly (ex: 'name', 'gReAt', '').
- [x] Make sure that queries like 'alskghadfljasdfjk' that return no results are handled properly.
- [x] Make sure that items of category textbook and non-textbook can be added successfully and then appear when searching for them or just selecting the category or clicking on the 'here' link.
- [x] Make sure that new textbooks can be added to the database by using the form.
- [x] Make sure that you cannot continue to the next page of the form if any required data is missing, or if any data is malformed
- [x] Make sure that when not logged in, the "create an item" link redirects to a login page
- [x] Make sure that when logged in as a user different from the one who created an item, when attempting to edit their item, you aren't allowed to
- [x] Make sure the manage page displays all of the created listings
- [x] Make sure that for each of the manageable listings, you can edit, archive, unarchive or delete it
- [x] Make sure that archived items let you know they're archived when you edit or view them
- [x] Make sure that archived items don't show up in search results or category view pages
- [x] Make sure that you can upload images from your computer and from an imgur link
- [x] Make sure that you can see the images on the sell confirmation page and on the listing view page
- [x] Make sure that once an item has been added, hitting "edit this item" and then continuing through the form doesn't change the details of the item


Makefile
```diff
@@ -13,7 +13,7 @@ lint:
 	yapf -i -r .
 
 test: init-test-db
-	python -m pytest .
+	python3 -m pytest .
```
dqu123: This shouldn't be necessary assuming your virtualenv is running.
Can you explain why you put this?

dqu123 (dismissed):

Makefile
tjleing: I guess my venv isn't working right for whatever reason.

tjleing: I'll try messing around with the venv folder and see if I can remove that line.

tjleing: Works now.

donut/modules/marketplace/helpers.py
```diff
@@ -1,30 +1,32 @@
...
 
 from donut.modules.core.helpers import get_member_data
+from donut.auth_utils import get_user_id, check_permission
+from donut.resources import Permissions
 
 from . import routes
 
 # taken from donut-legacy, which was apparently taken from a CS11
 # C++ assignment by dkong
 SKIP_WORDS = [
```
calebsander: Convert this to a `set` for constant-time lookup

donut/modules/marketplace/helpers.py
```diff
@@ -58,44 +54,259 @@ def render_with_top_marketplace_bar(template_url, **kwargs):
...
+    if num_cols == 0:
+        return flask.render_template('404.html'), 404
+
+    num_rows = ceil(num_cats / num_cols)
+
+    # Break categories into a 2d array so that it's easy to arrange in rows
+    # in the html file.
+    cats2d = [[]] * num_rows
+    for cat_index in range(len(categories)):
+        cats2d[cat_index // num_cols].append(categories[cat_index])
```
calebsander: Untested suggestion:
````python
cats2d = [categories[num_cols * row:num_cols * (row + 1)] for row in range(num_rows)]
````

File: donut/modules/marketplace/helpers.py [hunk 4]
```diff
@@ -58,44 +54,259 @@ def render_with_top_marketplace_bar(template_url, **kwargs):
...
+    return True
+
+
+def manage_delete_item(item_id):
+    """
+    Checks permissions and then deletes the item.
+    """
+    current_user_id = get_user_id(flask.session['username'])
+    if current_user_id != get_user_id_of_item(
+            item_id) and not check_permission(Permissions.ADMIN):
```
calebsander, line 78: Shouldn't this be an `or`? Or, `if not (current_user_id == ... and check_permission(...))`.

donut/modules/marketplace/helpers.py
```diff
@@ -58,44 +54,259 @@ def render_with_top_marketplace_bar(template_url, **kwargs):
...
+    field_index_map['edit'] = len(item) - 1
+
+    if item_active:
+        item.append('Archive')
+    else:
+        item.append('Unarchive')
+    field_index_map['archive'] = len(item) - 1
+
+    item.append('Delete')
+    field_index_map['delete'] = len(item) - 1
```
calebsander, line 218: If these lines are switched, wouldn't need the `- 1`. (Same for 2 earlier instances)

donut/modules/marketplace/helpers.py
```diff
@@ -58,44 +54,259 @@ def render_with_top_marketplace_bar(template_url, **kwargs):
...
+
+    links[field_index_map['edit']] = flask.url_for(
+        '.sell', item_id=item_id, state='edit')
+
+    if item_active:
+        links[field_index_map['archive']] = flask.url_for(
+            '.manage_confirm', item=item_id, state='archive')
+    else:
+        links[field_index_map['archive']] = flask.url_for(
+            '.manage_confirm', item=item_id, state='unarchive')
```
calebsander, line 258: Could use a ternary instead, since the left side is always the same

donut/modules/marketplace/helpers.py
```diff
@@ -104,427 +315,176 @@ def generate_search_table(fields=None, attrs={}, query=""):
...
-                    # so that we can use it to generate the links for each item
-                    item_id = int(data)
-                    temp_res_row.append(data)
+        #temp_res_row = item_listing
+        link_row = [''] * len(item_listing)
 
-                elif fields[field_index] == "item_timestamp":
-                    temp_res_row.append(data.strftime("%m/%d/%y"))
+        if 'item_id' in field_index_map:
+            item_id = int(item_listing[field_index_map['item_id']])
```
calebsander, line 339: Won't `item_id in field_index_map` always be `True`? If this assignment doesn't occur, the `item_id` value from the previous loop will be reused, which seems wrong.

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+
+    s = s_select_columns + s_from + s_where
+
+    # Execute the query.
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, list(attrs.values()))
+        result = cursor.fetchall()
+
+    # Return the list of lists.
+    result = list(map(lambda a: list(a.values()), result))
```
calebsander, line 693: `[list(a.values()) for a in result]`. Also, don't think you can assume the order of the values in each row `dict`; I would use `[[row[field] for field in fields] for row in result]`.

donut/modules/marketplace/helpers.py
```diff
@@ -104,427 +315,176 @@ def generate_search_table(fields=None, attrs={}, query=""):
...
-                    temp_res_row.append(get_name_from_user_id(int(data)))
+        if 'user_id' in field_index_map:
+            link_row[field_index_map['user_id']] = flask.url_for(
+                'directory_search.view_user',
+                user_id=int(item_listing[field_index_map['user_id']]))
 
-                elif fields[field_index] == "textbook_edition":
-                    temp_res_row.append(process_edition(data))
+            item_listing[field_index_map['user_id']] = get_name_from_user_id(
+                int(item_listing[field_index_map['user_id']]))
```
calebsander, line 351: Also, are all these `int` casts necessary? I thought `pymysqldb` returned `int`s already for `int` columns.

donut/modules/marketplace/helpers.py
```diff
@@ -104,427 +315,176 @@ def generate_search_table(fields=None, attrs={}, query=""):
...
+    item_index = field_index_map.get('item_title')
+    textbook_index = field_index_map.get('textbook_title')
 
     if item_index is None or textbook_index is None:
-        # can't merge, since the two columns aren't there
-        return (datalist, fields)
+        # Can't merge, since the two columns aren't there.
+        return (datalist, fields, field_index_map)
 
     for row_index in range(len(datalist)):
```
calebsander, line 407: `for row_index, row in enumerate(datalist):`

donut/modules/marketplace/helpers.py
```diff
@@ -104,427 +315,176 @@ def generate_search_table(fields=None, attrs={}, query=""):
...
         row = datalist[row_index]
-        if row[item_index] == "":
+        if row[item_index] == '':
             row[item_index] = row[textbook_index]
         del row[textbook_index]
         datalist[row_index] = row
```
calebsander, line 412: Don't think this is necessary; you never reassign to `row`, so it will already be the same value as `datalist[row_index]`

donut/modules/marketplace/helpers.py
```diff
@@ -538,24 +498,25 @@ def search_datalist(fields, datalist, query):
 
     for listing in datalist:
         item_tokens = []
+        is_isbn_match = False
         if get_category_name_from_id(
-                listing[field_index_map["cat_id"]]) == "Textbooks":
+                listing[field_index_map['cat_id']]) == 'Textbooks':
```
calebsander, line 479: Nit: might be more readable if you stored `listing[field_index_map['cat_id']]` in a variable so this condition wasn't split across 2 lines

donut/modules/marketplace/helpers.py
```diff
@@ -538,24 +498,25 @@ def search_datalist(fields, datalist, query):
...
-                listing[field_index_map["item_title"]])
-
-        # does the isbn match any of the query's isbns?
-        is_isbn_match = False
-        for isbn in query_isbns:
-            if listing[field_index_map["textbook_isbn"]] == isbn:
-                is_isbn_match = True
+                listing[field_index_map['item_title']])
 
         score = get_matches(query_tokens, item_tokens)
```
calebsander, line 497: This should go in the `else` clause of `if is_isbn_match`, since it isn't necessary to compute if `is_isbn_match` is `True`

donut/modules/marketplace/helpers.py
```diff
@@ -611,6 +557,8 @@ def get_matches(l1, l2):
     """
     Returns the number of matches between list 1 and list 2.
     """
+    l1 = set(l1)
+    l2 = set(l2)
     if len(l1) < len(l2):
         return len([x for x in l1 if x in l2])
```
calebsander, line 539: `len(l1 & l2)`

donut/modules/marketplace/helpers.py
```diff
@@ -621,28 +569,21 @@ def tokenize_query(query):
...
 
-    query = " ".join(query)
-    # Remove punctuation
-    punctuation = [",", ".", "-", "_", "!", ";", ":", "/", "\\"]
-    for p in punctuation:
-        query = query.replace(p, " ")
-    query = query.split()
+    # Remove punctuation.
+    query = ' '.join(query)
+    punctuation_regex = r'[,.\-_!;:/\\]'
```
calebsander, line 555: Maybe make this a global constant?

donut/modules/marketplace/helpers.py
```diff
@@ -661,28 +602,50 @@ def validate_isbn(isbn):
...
+    isbn = isbn.replace('-', '')
+
+    # Regexes shamelessly copypasted.
+    # The ISBN-10 can have an x at the end (but the ISBN-13 can't).
+    if re.match('^[0-9]{9}[0-9x]$', isbn, re.IGNORECASE) != None:
+        # Check the check digit.
+        total = 0
+        for i in range(10):
+            char = isbn[i].lower()
+            digit = 10  # x has value 10.
```
calebsander, line 596:
````python
for i, char in enumerate(isbn):
    digit = 10 if char.lower() == 'x' else int(char)
    # ...
````

File: donut/modules/marketplace/helpers.py [hunk 17]
```diff
@@ -661,28 +602,50 @@ def validate_isbn(isbn):
...
+            total += digit * weight
+        return total % 11 == 0
+
+    if re.match('^[0-9]{13}$', isbn, re.IGNORECASE) != None:
+        # Check the check digit.
+        total = 0
+        for i in range(13):
+            weight = 1
+            if i % 2 != 0:
+                weight = 3
```
calebsander, line 609: `weight = 1 if i % 2 == 0 else 3`

donut/modules/marketplace/helpers.py
```diff
@@ -661,28 +602,50 @@ def validate_isbn(isbn):
...
+            digit = int(isbn[i])
+            total += digit * weight
+        if total % 10 == 0:
+            return True
+        else:
+            return False
```
calebsander, line 615: `return total % 10 == 0`

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
-                return str(edition) + "st"
+                return str(edition) + 'st'
             if edition % 10 == 2:
-                return str(edition) + "nd"
+                return str(edition) + 'nd'
             if edition % 10 == 3:
-                return str(edition) + "rd"
-            return str(edition) + "th"
+                return str(edition) + 'rd'
+            return str(edition) + 'th'
```
calebsander, line 647: Rather than converting `edition` to `int` and then back to `str`, could just store the `int` in a separate variable

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
         else:
             return str(edition)
     except ValueError:
```
calebsander, line 650: No need to wrap all this code in a `try`-`except`; could just wrap the `edition = int(edition)` line

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
         return edition
 
 
+#################
+# TABLE QUERIES #
+#################
+def get_table_list_data(tables, fields=None, attrs={}):
```
calebsander, line 657: Or default `fields` to `['*']` and you wouldn't need the special case

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+                are used.
+        attrs:  The attributes of the members to filter for.
+    Returns:
+        result: The fields and corresponding values of members with desired
+                attributes. In the form of a list of lists.
+    """
+    if fields == None:
+        s_select_columns = 'SELECT * '
+    else:
+        s_select_columns = 'SELECT ' + ', '.join(fields) + ' '
```
calebsander, line 676: Move the trailing space to the start of `s_from` for consistency

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+
+    s_from = 'FROM ' + tables
+
+    s_where = ''
```
calebsander, line 680: Can go in `else` clause of `if attrs`

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+    s = s_select_columns + s_from + s_where
+
+    # Execute the query.
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, list(attrs.values()))
+        result = cursor.fetchall()
+
+    # Return the list of lists.
+    result = list(map(lambda a: list(a.values()), result))
+    return result
```
calebsander, line 694: Could just `return` rather than assigning to a temporary variable

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+
+
+def table_fetch_one(tables, fields=None, attrs={}):
```
calebsander, line 697: This function almost completely duplicates `get_table_list_data()` -- would be better to write one function to handle their  common functionality

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+    result = list(result.values())
+
+    # If only one field requested, unwrap [_] to _
+    if len(fields) == 1:
+        result = result[0]
+
+    return result
+
+
+def table_fetch_all(tables, fields=None, attrs={}):
```
calebsander, line 742: This function could easily be rewritten to call `get_table_list_data()`. I think the only difference is the special case for when `fields` is a singleton.

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, list(attrs.values()))
+        result = cursor.fetchall()
+
+    # Return the list of lists
+    result = list(map(lambda a: list(a.values()), result))
+
+    # If only one field requested, unwrap [[_], [_], [_]] to [_, _, _]
+    if len(fields) == 1:
+        result = list(map(lambda a: a[0], result))
```
calebsander, line 782:
```suggestion
        result = [a[0] for a in result]
```

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+        'cat_id', 'item_title', 'item_condition', 'item_details', 'item_price',
+        'textbook_id', 'textbook_edition', 'textbook_isbn', 'state'
+    ]
+
+    to_return = []
+    for parameter in parameters:
+        if parameter in skip_fields:
+            continue
+        if parameter in flask.request.form:
+            to_return.append([parameter, flask.request.form[parameter]])
```
calebsander, line 811: Use a tuple here instead of a 2-element list?

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+        category_id = int(flask.request.form['cat_id'])
+    except ValueError:
+        # this should never happen through normal form use
+        errors.append('Somehow the category got all messed up.')
+
+    cat_title = get_table_list_data(
+        'marketplace_categories',
+        fields=['cat_title'],
+        attrs={'cat_id': category_id})
+    if len(cat_title) == 0:
```
calebsander, line 839: `if cat_title` (and switch the `if` and `else` clauses)

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+        errors.append('Somehow the category got all messed up.')
+    else:
+        # 2d array to a single element
+        # [['Furniture']] -> 'Furniture'
+        cat_title = cat_title[0][0]
+
+    if cat_title == 'Textbooks':
+        # regex to make sure the textbook edition is valid
+        if 'textbook_edition' in flask.request.form:
+            edition_regex = '^(|[0-9]+|international)$'
```
calebsander, line 850: Could use `*` instead of `+` and remove the empty string case. Also, can this be made a global constant?

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+            if re.match(edition_regex,
+                        str(flask.request.form['textbook_edition']),
+                        re.IGNORECASE) == None:
+                errors.append(
+                    'Textbook edition is invalid. Try providing a number or \'International\'.'
```
calebsander, line 855: Use double-quotes around the string to avoid having to escape the single-quotes

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+
+    # condition is mandatory
+    if not 'item_condition' in flask.request.form or flask.request.form['item_condition'] == '':
+        errors.append('Item condition must be set.')
+
+    # price must be present and valid
+    if not 'item_price' in flask.request.form or flask.request.form['item_price'] == '':
+        errors.append('Price cannot be left blank.')
+    else:
+        price_regex = '^([0-9]{1,4}\.[0-9]{0,2}|[0-9]{1,4}|\.[0-9]{1,2})$'
```
calebsander, line 878: Can combine `[0-9]{1,4}\.[0-9]{0,2}|[0-9]{1,4}` into `[0-9]{1,4}(\.[0-9]{0,2})?`. Also, you should really use a regex string literal. It is not obvious to me that `'\.' == '\\.'` (in C and JavaScript, for instance, that is not the case).

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+            )
+
+    image_regex_1 = "^https?://i\.imgur\.com/[a-z0-9]+\.(jpg|png|gif)$"
+    image_regex_2 = "^https?://imgur\.com/[a-z0-9]+$"
+    for image in flask.request.form.get('item_images', []):
+        if image == "":
+            continue
+
+        if re.match(image_regex_1, image, re.IGNORECASE) == None:
+            if re.match(image_regex_2, image, re.IGNORECASE) != None:
```
calebsander, line 903: Would be more readable if you negated this condition (and switched the clauses) or removed the `!= None`

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+    Returns:
+        the item_id, or -1 if it fails
+    """
+    user_id = int(stored['user_id'])
+    cat_id = int(stored['cat_id'])
+    cat_title = stored['cat_title']
+    item_condition = stored['item_condition']
+    item_details = stored['item_details']
+    item_price = stored['item_price']
+    item_images = []
```
calebsander, line 932: This statement is unnecessary

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+            item_id = cursor.lastrowid
+    else:
+        item_title = stored['item_title']
+        s = '''INSERT INTO marketplace_items (user_id, cat_id, item_title, item_condition, item_details, item_price) VALUES (%s, %s, %s, %s, %s, %s)'''
+        with flask.g.pymysql_db.cursor() as cursor:
+            cursor.execute(s, [
+                user_id, cat_id, item_title, item_condition, item_details,
+                item_price
+            ])
+            item_id = cursor.lastrowid
```
calebsander, line 957: These two clauses share a bit of duplicated code

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+
+    if item_id == -1:
+        return -1
+
+    s = 'INSERT INTO marketplace_images (item_id, img_link) VALUES (%s, %s);'
```
calebsander, line 962:
```suggestion
    s = 'INSERT INTO marketplace_images (item_id, img_link) VALUES (%s, %s)'
```
for consistency

donut/modules/marketplace/helpers.py
```diff
@@ -693,25 +656,397 @@ def process_edition(edition):
...
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, [item_id])
+
+    s = 'INSERT INTO marketplace_images (item_id, img_link) VALUES (%s, %s);'
+    for image in item_images:
+        if image == "":
+            continue
+        with flask.g.pymysql_db.cursor() as cursor:
+            cursor.execute(s, [item_id, image])
+    return item_id
```
calebsander, line 1023: This function duplicates most of the code in `create_new_listing()`. Would be nice to factor out their common functionality

donut/modules/marketplace/helpers.py
```diff
@@ -726,21 +1061,72 @@ def add_textbook(title, author):
...
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, [item_id])
+        result = cursor.fetchone()
+    if result == None:
+        # the item doesn't exist
+        return False
+    return result['user_id']
+
+
 def get_name_from_user_id(user_id):
```
calebsander, line 1106: Could instead use `core.helpers.get_name_and_email()`

donut/modules/marketplace/helpers.py
```diff
@@ -768,13 +1155,15 @@ def get_textbook_info_from_textbook_id(textbook_id):
...
+    s = 'SELECT textbook_title, textbook_author FROM marketplace_textbooks WHERE textbook_id=%s'
+
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, [textbook_id])
+        result = cursor.fetchone()
+
     if result == None:
         return None
-    return list(result)
+    return [result['textbook_title'], result['textbook_author']]
```
calebsander, line 1142: Tuples allow for better optimization:
```suggestion
    return (result['textbook_title'], result['textbook_author'])
```

donut/modules/marketplace/helpers.py
```diff
@@ -786,10 +1175,11 @@ def get_category_name_from_id(cat_id):
...
-    result = flask.g.db.execute(query, cat_id=cat_id).first()
+    s = 'SELECT cat_title FROM marketplace_categories WHERE cat_id=%s'
+
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, [cat_id])
+        result = cursor.fetchone()
     if result == None:
         return None
-    return result[0]
+    return result['cat_title']
```
calebsander, line 1161: `return result and result['cat_title']`

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
+        'textbook_author'
     ]
     data = helpers.get_table_list_data(
-        ['marketplace_items',
-         'marketplace_textbooks'], stored_fields, {'item_id': item_id})[0]
+        'marketplace_items NATURAL LEFT JOIN marketplace_textbooks',
+        stored_fields, {'item_id': item_id})
+
+    # make sure the item_id is a valid item, i.e. data is nonempty
+    if len(data) == 0:
```
calebsander, line 131:
```suggestion
    if not data:
```

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
+    if not stored['item_active']:
+        flask.flash('The listing for this item is no longer active!')
+
+    # grab the stored image links
+    data = helpers.get_table_list_data('marketplace_images', ['img_link'],
+                                       {'item_id': item_id})
+    # unwrap elements of 2d array
+    stored_images = list(map(lambda x: x[0], data))
+    # pad out to MAX_NUM_IMAGES
+    stored_images += [""] * (MAX_NUM_IMAGES - len(stored_images))
```
calebsander, line 167:
```suggestion
    stored_images += [''] * (MAX_NUM_IMAGES - len(stored_images))
```

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
+
+    # notify if the item is inactive
+    if not stored['item_active']:
+        flask.flash('The listing for this item is no longer active!')
+
+    # grab the stored image links
+    data = helpers.get_table_list_data('marketplace_images', ['img_link'],
+                                       {'item_id': item_id})
+    # unwrap elements of 2d array
+    stored_images = list(map(lambda x: x[0], data))
```
calebsander, line 165:
```suggestion
    stored_images = [x[0] for x in data]
```

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
 
 @blueprint.route('/marketplace/sell', methods=['GET', 'POST'])
 def sell():
+    if 'username' not in flask.session:
+        # they're not logged in, kick them out
+        flask.flash('Login required to access that page.')
+        flask.session['next'] = flask.url_for('.sell')
+        return helpers.render_with_top_marketplace_bar('requires_login.html')
+
+    from donut.modules.marketplace.constants import Page
```
calebsander, line 255: Conditional imports :(

donut/modules/marketplace/routes.py
```diff
@@ -166,42 +278,78 @@ def sell():
...
-            data = helpers.get_table_list_data([
-                'marketplace_items', 'marketplace_textbooks'
-            ], stored_fields, {'item_id': item_id})[0]
-            for i in range(len(data)):
-                stored[stored_fields[i]] = data[i]
+            data = helpers.get_table_list_data(
+                'marketplace_items NATURAL LEFT JOIN marketplace_textbooks',
+                stored_fields, {'item_id': item_id})
+
+            if len(data) == 0:
```
calebsander, line 313: `if data:` and switch clauses

donut/modules/marketplace/routes.py
```diff
@@ -166,42 +278,78 @@ def sell():
...
+            else:
+                # unwrap 2d array
+                data = data[0]
+                for i in range(len(data)):
+                    stored[stored_fields[i]] = data[i]
+
+            # make sure the current user is the one who posted the item
+            current_user_id = get_user_id(flask.session['username'])
+            if stored['user_id'] != current_user_id and not check_permission(
+                    Permissions.ADMIN):
```
calebsander, line 327: This check is used all over the place -- can it be factored into a separate function?

donut/modules/marketplace/routes.py
```diff
@@ -166,42 +278,78 @@ def sell():
...
+    # images
+    if 'item_images[]' in flask.request.form:
+        stored_images = flask.request.form.getlist('item_images[]')
+    else:
+        data = helpers.get_table_list_data('marketplace_images', ['img_link'],
+                                           {'item_id': item_id})
+        # unwrap elements of 2d array
+        stored_images = list(map(lambda x: x[0], data))
+        # pad out to MAX_NUM_IMAGES
+        stored_images += [""] * (MAX_NUM_IMAGES - len(stored_images))
```
calebsander, line 346:
```suggestion
        stored_images += [''] * (MAX_NUM_IMAGES - len(stored_images))
```

donut/modules/marketplace/routes.py
```diff
@@ -166,42 +278,78 @@ def sell():
...
+                        flask.request.referrer)  # go back to the previous page
+
+    # images
+    if 'item_images[]' in flask.request.form:
+        stored_images = flask.request.form.getlist('item_images[]')
+    else:
+        data = helpers.get_table_list_data('marketplace_images', ['img_link'],
+                                           {'item_id': item_id})
+        # unwrap elements of 2d array
+        stored_images = list(map(lambda x: x[0], data))
```
calebsander, line 344:
```suggestion
        stored_images = [x[0] for x in data]
```

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
+    # Page.TEXTBOOK:     Select a textbook (special, appears between pages 1 and 2 if the category is 'Textbooks')
+    # Page.INFORMATION:  Input information about the listing
+    # Page.CONFIRMATION: Confirm that info is correct
+    # Page.SUBMIT:       Add data to database, show success message
+
+    page = Page.CATEGORY  # default is first page; category select page
+    if 'page' in flask.request.form:
         # but if we pass it in, get it
-        page = int(flask.request.form["page"])
+        page = Page.__members__[flask.request.form['page']]
```
calebsander, line 270:
```suggestion
        page = Page[flask.request.form['page']]
```

donut/modules/marketplace/routes.py
```diff
@@ -86,76 +97,177 @@ def query():
...
 
-    if not page in [1, 10, 2, 3, 4]:
+    if not page in Page:
```
calebsander, line 272: Can't check this here; if `page` is not a valid page, `Page.__members__[flask.request.form['page']]` will throw a key error

donut/modules/marketplace/routes.py
```diff
@@ -210,38 +358,38 @@ def sell():
...
             else:
-                textbook_title = flask.request.form["textbook_title"]
-                textbook_author = flask.request.form["textbook_author"]
-                if textbook_title == "" or textbook_author == "":
+                textbook_title = flask.request.form['textbook_title']
+                textbook_author = flask.request.form['textbook_author']
+                if textbook_title == '' or textbook_author == '':
                     flask.flash(
-                        "Textbook title and textbook author can't be blank.")
+                        'Textbook title and textbook author can\'t be blank.')
```
calebsander, line 394: Double-quoted is nicer to avoid escaping the single quote

donut/modules/marketplace/routes.py
```diff
@@ -253,57 +401,61 @@ def sell():
...
             # if the user was on page 1 and hit continue, or on page 2 and hit back, send them to 10 instead
-            if (page == 2 and prev_page == 1) or (page == 1
-                                                  and prev_page == 2):
-                page = 10
-
-    if page >= 2 and page <= 4 and cat_title == "Textbooks":
-        if "textbook_id" not in flask.request.form or flask.request.form["textbook_id"] == "":
-            flask.flash("You need to select a textbook.")
+            if (page == Page.INFORMATION and prev_page == Page.CATEGORY) or (
+                    page == Page.CATEGORY and prev_page == Page.INFORMATION):
```
calebsander, line 410: `if set([page, prev_page]) == set([Page.CATEGORY, page.INFORMATION]):`

donut/modules/marketplace/routes.py
```diff
@@ -253,57 +401,61 @@ def sell():
...
+                page = Page.TEXTBOOK
+
+    if page in [Page.INFORMATION, Page.CONFIRMATION, Page.SUBMIT
```
calebsander, line 413: `set` pls

donut/modules/marketplace/routes.py
```diff
@@ -313,20 +465,20 @@ def sell():
...
-                     "textbook_author"], {"textbook_id": textbook_id})
-                (stored["textbook_title"],
-                 stored["textbook_author"]) = textbook_info[0]
+                    'marketplace_textbooks',
+                    ['textbook_title',
+                     'textbook_author'], {'textbook_id': textbook_id})
+                (stored['textbook_title'],
+                 stored['textbook_author']) = textbook_info[0]
             if field in flask.request.form:
                 stored[field] = flask.request.form[field]
```
calebsander, line 485: `stored.update(flask.request.form)`

donut/modules/marketplace/routes.py
```diff
@@ -376,68 +531,49 @@ def sell():
...
+        stored['user_id'] = get_user_id(flask.session['username'])
+        stored['item_images'] = stored_images
+
         # if we are creating a new item listing, insert it into the database.
         # otherwise, if we are editing an item that already exists, we need
         # to update the listing.
-        stored["cat_title"] = cat_title
-        if (state == "new"):
+        stored['cat_title'] = cat_title
+        if (state == 'new'):
```
calebsander, line 575:
```suggestion
        if state == 'new':
```

donut/modules/marketplace/templates/manage/confirm.html
```diff
@@ -0,0 +1,43 @@
...
+            {% endfor %}
+        </tr>
+    </thead>
+    <tbody>
+        <!-- display item to delete as a table -->
+        <tr>
+            {% for item_field in item %}
+                {% set i = loop.index0 %}
+                <td>
+                    {% if links[i] %} <a href={{ links[i] }}> {% endif %}
```
calebsander, line 27:
```suggestion
                    {% if links[i] %} <a href="{{ links[i] }}"> {% endif %}
```

donut/modules/marketplace/templates/manage/confirm.html
```diff
@@ -0,0 +1,43 @@
...
+                    {{ item_field }}
+                    {% if links[i] %} </a> {% endif %}
+                </td>
+            {% endfor %}
+        </tr>
+    </tbody>
+</table>
+
+<form action={{ url_for(".manage") }} method="post">
```
calebsander, line 36:
```suggestion
<form action="{{ url_for(".manage") }}" method="POST">
```

donut/modules/marketplace/templates/manage/manage.html
```diff
@@ -0,0 +1,87 @@
...
+            <tbody>
+                <!-- display data in one main table -->
+                {% for row in activelist %}
+                    {% set i = loop.index0 %}
+                    <tr>
+                        {% for cell in row %}
+                            {% set j = loop.index0 %}
+                            {% set link = activelinks[i][j] %}
+                            <td>
+                                {% if link %} <a href={{ link }}> {% endif %}
```
calebsander, line 32:
```suggestion
                                {% if link %} <a href="{{ link }}"> {% endif %}
```

donut/modules/marketplace/templates/manage/manage.html
```diff
@@ -0,0 +1,87 @@
...
+            <tbody>
+                <!-- display data in one main table -->
+                {% for row in inactivelist %}
+                    {% set i = loop.index0 %}
+                    <tr>
+                        {% for cell in row %}
+                            {% set j = loop.index0 %}
+                            {% set link = inactivelinks[i][j] %}
+                            <td>
+                                {% if link %} <a href={{ link }}> {% endif %}
```
calebsander, line 71:
```suggestion
                                {% if link %} <a href="{{ link }}"> {% endif %}
```

donut/modules/marketplace/templates/requires_login.html
```diff
@@ -0,0 +1,9 @@
+{% extends "layout.html" %}
+{% block page %}
+
+<p class="login-info">
+    This page requires you to login with your username and password.
+    Login <a href={{ url_for("auth.login") }}>here</a> and you'll be redirected.
```
calebsander, line 6:
```suggestion
    Login <a href="{{ url_for("auth.login") }}">here</a> and you'll be redirected.
```

donut/modules/marketplace/templates/search.html
```diff
@@ -19,13 +19,16 @@ <h4>
...
                     <tr>
-                        {% for j in range(datalist[i]|length) %}
+                        {% for cell in row %}
+                            {% set j = loop.index0 %}
+                            {% set link = links[i][j] %}
                             <td>
-                                {% if links[i][j] != 0 %} <a href={{ links[i][j] }}> {% endif %}
-                                {{ datalist[i][j] }}
-                                {% if links[i][j] != 0 %} </a> {% endif %}
+                                {% if link %} <a href={{ link }}> {% endif %}
```
calebsander, line 29:
```suggestion
                                {% if link %} <a href="{{ link }}"> {% endif %}
```

donut/modules/marketplace/templates/sell/sell_1.html
```diff
@@ -18,18 +18,20 @@ <h3>Modify an existing item</h3>
...
             <select class="form-control" name="cat_id" id="cat_id" autofocus>
-                {% for i in range(cats|length) %}
-                    {% for j in range(cats[i]|length) %}
+                {% for cat_row in cats %}
+                    {% set i = loop.index0 %}
+                    {% for cat in cat_row %}
+                        {% set j = loop.index0 %}
                         <option value="{{ i*cats[0]|length+j+1 }}"
-                        {% if i*cats[0]|length+j+1 == category_id %} selected {% endif %} >
+                        {% if i*cat_row|length+j+1 == category_id %} selected {% endif %} >
```
calebsander, line 26: Would be cleaner to set some variable to `i*cat_row|length+j+1`. Also, please put parens around `cat_row|length`.

donut/modules/marketplace/templates/sell/sell_10.html
```diff
@@ -55,7 +57,11 @@ <h3>Modify an existing item</h3>
     <br>
     <br>
     <p>Alternatively, if you can't find your textbook, you can add it here.</p>
-    <form action={{ url_for(".sell", state=state) }} method="post">
+    {% if state=='edit' %}
+    <form action={{ url_for(".sell", state=state, item_id=request.args['item_id']) }} method='post'>
+    {% else %}
+    <form action={{ url_for(".sell", state=state) }} method='post'>
+    {% endif %}
```
calebsander, line 64: These can be collapsed into a single `form` element. Use a ternary in the `action`.

donut/modules/marketplace/templates/sell/sell_2.html
```diff
@@ -77,10 +77,40 @@ <h3>Modify an existing item</h3>
...
+                {% endfor %}
+            </table>
+        </div>
+
+        <script>
+            var IMGUR_CLIENT_ID = '{{ imgur_id }}';
+            var LOADING_PATH = "{{ url_for('static', filename='images/marketplace/loading.gif') }}";
+        </script>
+        <script src={{ url_for('static', filename='js/marketplace/overlay.js') }}></script>
+        <script src={{ url_for('static', filename='js/marketplace/sell_images.js') }}></script>
```
calebsander, line 108: It's a good habit to wrap HTML attribute values in quotes

donut/modules/marketplace/templates/topbar.html
```diff
@@ -1,29 +1,31 @@
...
-                            </td>
-                            {% endfor %}
-                        </tr>
+                        {% for row in cats %}
+                            {% set i = loop.index0 %}
+                            <tr>
+                                {% for cat in row %}
+                                    {% set j = loop.index0 %}
+                                    <td style="width: {{ width }}%">
+                                        <a href={{ url_for(".category", cat=i*row|length+j+1) }}>{{ cat }}</a>
```
calebsander:
```suggestion
                                        <a href="{{ url_for(".category", cat=i*(row|length)+j+1) }}">{{ cat }}</a>
```

donut/modules/marketplace/templates/topbar.html
```diff
@@ -45,11 +47,13 @@ <h2>Marketplace</h2>
...
-                                {% for j in range(cats[i]|length) %}
-                                <option {% if i*cats[0]|length+j+1 == cat_id %} selected {% endif%}
-                                        value={{ i*cats[0]|length+j+1 }}>
-                                    {{ cats[i][j] }}
+                            {% for row in cats %}
+                                {% set i = loop.index0 %}
+                                {% for cat in row %}
+                                    {% set j = loop.index0 %}
+                                <option {% if i*row|length+j+1 == cat_id %} selected {% endif%}
+                                        value={{ i*row|length+j+1 }}>
```
calebsander, line 43: Store `i*row|length+j+1` in a variable and use parens please

donut/modules/marketplace/templates/view_item.html
```diff
@@ -11,8 +11,11 @@ <h3>
         {{stored["item_title"]}}
         {% endif %}
     </h3>
-
-    <!-- TODO: when logins and privs are implemented, put "edit this item" here -->
+    {% if has_edit_privs %}
+    <h4>
+        <a href={{ url_for(".sell", state='edit', item_id=request.args["item_id"]) }}>Edit this listing</a>
```
calebsander, line 16:
```suggestion
        <a href="{{ url_for('.sell', state='edit', item_id=request.args['item_id']) }}">Edit this listing</a>
```

donut/static/css/marketplace/overlay.css
```diff
@@ -0,0 +1,53 @@
...
+
+#overlayBackground {
+    background-color: #000000;
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    opacity: 0.7;
+    z-index: 9000;
```
calebsander, line 11: ew

donut/static/js/marketplace/overlay.js
```diff
@@ -0,0 +1,73 @@
...
+function showOverlay()
+{
+    // Check if an overlay already exists.
+    if ($("#overlay").length)
+        return;
+
+    // Create the overlay. The overlayBackground and overlayWindow are child
+    // elements of the overlay div.
+    var overlay = document.createElement("div");
+    overlay.setAttribute("id", "overlay");
```
calebsander, line 29: Can do stuff like `var overlay = $('<div>').attr('id', 'overlay')` in jQuery

donut/static/js/marketplace/overlay.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/*
+ * Centers overlayWindow. This should be called whenever the size of the
+ * overlay window changes.
+ */
+function centerOverlay()
+{
+    var width = $("#overlayWindow").width();
+    var height = $("#overlayWindow").height();
```
calebsander, line 62: Store `$("#overlayWindow")` in a variable to avoid computing it twice

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+
+/*
+ * Display the form to upload an image.
+ */
+function showUploadForm(imageNum)
+{
+  showOverlay();
+
+  // Create the forms to upload images.
+  var uploadForm = "<div style=\"margin: 10px\">";
```
calebsander, line 20: Single quotes here and later would allow you to not escape the HTML quotes

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+ * to imgur using the sell form. This requires the overlay functionality
+ * found in overlay.js.
+ */
+
+//-------------------------------------------------------------------
+
+/*
+ * Display the form to upload an image.
+ */
+function showUploadForm(imageNum)
```
calebsander, line 15: I don't know the code well enough to answer this, but are we SURE `imageNum` always comes from a secure source? Otherwise, there is some nasty XSS potential here since you are manually building HTML strings.

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+  uploadForm += "</div>";
+
+  // Display the form.
+  $("#overlayWindow").append(uploadForm);
+
+  centerOverlay();
+
+  // Set an error handler for ajax errors. This way, we let imgur handle
+  // invalid urls or files that the user may try to upload.
+  $(document).ajaxError(function() {
```
calebsander, line 57: `$(document).ajaxError(handleImageUploadError)`

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+  // Create the forms to upload images.
+  var uploadForm = "<div style=\"margin: 10px\">";
+
+  // Used to report error messages.
+  uploadForm += "<div id=\"uploadImageError\"></div>";
+  uploadForm += "<table id=\"uploadImageForm\">";
+  uploadForm += "<tr>";
+  uploadForm += "<td>Upload from your computer:</td>";
+  uploadForm += "<td>Upload from the web:</td>";
+  uploadForm += "</tr><tr>";
```
calebsander, line 28: Rather than `+=`ing each thing, just add a `+` at the end of the string on each line

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+  var reader = new FileReader();
+
+  // Make sure a file was selected.
+  if ($("#image_file").prop("files").length === 0)
+  {
+    handleImageUploadError("You must select a file first.")
+    return false;
+  }
+
+  var imageFile = $("#image_file").prop("files")[0];
```
calebsander, line 132: Store `$("#overlayWindow")` in a variable

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+
+  // Set the file reader so that it calls the upload image function when
+  // the read is finished.
+  reader.onload = function(e) {
+    // Convert binary data to base 64 and then upload it.
+    uploadImage(btoa(e.target.result), imageNum, "base64");
```
calebsander, line 138:
```suggestion
    uploadImage(btoa(reader.result), imageNum, "base64");
```

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+    },
+    data: {
+      image: imageData,
+      type: uploadType,
+    },
+    // Forces jQuery to parse the response properly.
+    dataType: "json",
+    success: function(response) {
+      // Call the event handler with an additional argument.
+      console.log(response);
```
calebsander, line 163: Remove after debugging is done

donut/static/js/marketplace/sell_images.js
```diff
@@ -0,0 +1,219 @@
...
+  });
+}
+
+/*
+ * This function handles the response from imgur on a successful request.
+ */
+function handleUploadResponse(response, imageNum)
+{
+  // Check that the request was actually successful.
+  if (response.status != 200 || response.success != true)
```
calebsander, line 175:
```suggestion
  if (response.status != 200 || !response.success)
```

sql/test_data.sql
```diff
@@ -58,6 +58,13 @@ INSERT INTO member_options(user_id, option_id, option_type) VALUES
 INSERT INTO images(user_id, extension, image) VALUES
     (3, 'png', 'NOT_A_REAL_IMAGE');
 
+INSERT INTO marketplace_categories(cat_title, cat_order) VALUES
+    ('Furniture', 1),
+    ('Textbooks', 2);
+
+INSERT INTO marketplace_items(cat_id, user_id, item_title, item_details, item_condition, item_price) VALUES
+    (1, 1, 'A table', "It\'s a pretty cool table, I guess", 'Good', 5.99);
```
calebsander, line 66: Don't think `\` is necessary, but not sure

tests/modules/marketplace/test_marketplace_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+    result = get_table_list_data(table, attrs=attrs)
+    assert len(result) == 1  # only one result
+
+    table = 'marketplace_items'
+    fields = ['item_price', 'item_title']
+    attrs = {'cat_id': 1}
+    result = get_table_list_data(table, fields=fields, attrs=attrs)
+    assert len(result) == 1  # only one result
+    assert len(result[0]) == 2  # result has two fields
+    assert result[0] == [Decimal('5.99'), 'A table']
```
calebsander, line 32: Combine into single assertion: `assert result == [[Decimal('5.99'), 'A table']]`

tests/modules/marketplace/test_marketplace_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+    query = "and,if great testing testing ,.,..,/./,/,./,./,"
+    assert tokenize_query(query) == ['great', 'testing', 'testing']
+
+
+def test_validate_isbn(client):
+    assert validate_isbn("978-1-60309-419-1") == True
+    assert validate_isbn("9971-5-0210-0") == True
+    assert validate_isbn("0-8044-2957-X") == True
+    assert validate_isbn("0-9752298-0-X") == True
+    assert validate_isbn("978-3-16-148410-0") == True
```
calebsander, line 60: `== True` is not needed

tests/modules/marketplace/test_marketplace_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+    assert validate_isbn("978-1-60309-419-0") == False
+    assert validate_isbn("9971-5-0210-X") == False
+    assert validate_isbn("0-8044-2957-1") == False
+    assert validate_isbn("0-9752298-0-2") == False
+    assert validate_isbn("978-3-16-148400-0") == False
```
calebsander, line 66: Likewise, `== False` can be removed and `not` can be added to the front

tests/modules/marketplace/test_marketplace_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+    assert process_edition(2.0) == "2nd"
+    assert process_edition(3.0) == "3rd"
+    assert process_edition(4.0) == "4th"
+    assert process_edition(5.0) == "5th"
+    assert process_edition(6.0) == "6th"
+    assert process_edition(7.0) == "7th"
+    assert process_edition(8.0) == "8th"
+    assert process_edition(9.0) == "9th"
+    assert process_edition(10.0) == "10th"
+    assert process_edition(2017.0) == "2017"
```
calebsander, line 80: Are floats needed?

tests/modules/marketplace/test_marketplace_routes.py
```diff
@@ -0,0 +1,92 @@
...
+    with client as c:
+        with c.session_transaction() as sess:
+            sess['username'] = 'username'
+
+        # this with block is necessary to keep the client around,
+        # as we can only dig into flask.request[] during the actual
+        # request (see http://flask.pocoo.org/docs/0.12/testing/#keeping-the-context-around)
+
+        rv = c.post(
+            flask.url_for('marketplace.sell'), data=dict(page='CATEGORY'))
```
calebsander, line 67: Why not use a dict literal? Same below

calebsander (changes requested):

donut/modules/marketplace/helpers.py
tjleing: No need anymore -- using flexbox to do it frontend now.  Getting rid of this old code feels good

donut/modules/marketplace/helpers.py
tjleing: I think it's right -- if the person trying to make the change is neither the owner of the item nor an admin, don't make it.  I'll agree that it's a little confusing; do you think `if not (current_user_id == ... or check_permission(...))` would be better?

donut/modules/marketplace/helpers.py
tjleing: Hmm, this only works if fields isn't `None`; above, if fields is `None` I use a `SELECT *`.  I think I'll use both of your suggestions -- `return [list(row.values()) for row in result]` if fields is `None` and `return [[row[field] for field in fields] for row in result]` otherwise.

donut/modules/marketplace/helpers.py
tjleing: whoa

donut/modules/marketplace/helpers.py
tjleing: I would use the cool apply suggestion button, but I wanted to combine all these changes into one commit if possible.

donut/modules/marketplace/helpers.py
calebsander: Yeah, I definitely misread this the first time I saw it. I figured it out later but guess I forgot to remove this comment. I think `if not (... or ...):` is a little more readable since it suggests you are doing something only in the exceptional case that the user is *not* allowed to do this. But totally your call.

donut/modules/marketplace/helpers.py
calebsander: Or, if `fields` was originally `None`, let `fields = list(result[0]) if result else []`

donut/modules/marketplace/constants.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class Page(Enum):
```
calebsander, line 4: Use `IntEnum` instead?

PR #45, opened 2017-11-09
Added Singular Group Endpoint

majesticfish:
### Summary
- Added a method in helpers to get the data of a singular group. Added the endpoint in the routes file in the groups modules
### Test Plan
- go to `localhost:9000/1/groups/` to see all groups. Select one and go to `localhost:9000/1/groups/<group-id>`. This should show the information on the particular group.

### DIFF
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -60,11 +60,11 @@ def get_group_data(group_id, fields=None):
         "members_can_send", "newsgroups", "visible", "admin_control_members"
     ]
     default_fields = ["group_id", "group_name", "group_desc", "type"]
-    if fields == None:
+    if fields is None:
         fields = default_fields
     else:
         if any(f not in all_returnable_fields for f in fields):
-            return "Invalid field"
+            return None 
 
     # Build the SELECT and FROM clauses
     s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("groups"))
@@ -74,6 +74,10 @@ def get_group_data(group_id, fields=None):
 
     # Execute the query
     result = flask.g.db.execute(s, g=group_id).first()
+   
+    # Check to see if query returned anything
+    if result is None:
+        return {}
 
     # Return the row in the form of a dict
     result = {f: t for f, t in zip(fields, result)}
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -64,7 +64,7 @@ def get_group_data(group_id, fields=None):
         fields = default_fields
     else:
         if any(f not in all_returnable_fields for f in fields):
-            return None 
+            return None
 
     # Build the SELECT and FROM clauses
     s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("groups"))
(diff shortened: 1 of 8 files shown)


dqu123:
@majesticfish, did you delete the test plan section from the PR comment? You can put a link to the endpoint (i.e. `localhost:9000/1/groups/1/` should show ...) to make it easier for the reviewer
to run and test your code.

donut/modules/groups/routes.py
```diff
@@ -16,3 +16,9 @@ def get_groups_list():
     if "fields" in flask.request.args:
         fields = [f.strip() for f in flask.request.args["fields"].split(',')]
     return json.dumps(helpers.get_group_list_data(fields=fields, attrs=attrs))
+
+
+@blueprint.route("/1/groups/<int:group_id>/")
+def get_groups(group_id):
+    """GET /1/groups/<int:group_id>/"""
+    return json.dumps(helpers.get_group_data(group_id))
```
dqu123: We should use `jsonify` (`from flask import jsonify`). It wraps `json.dumps`.
See http://flask.pocoo.org/docs/0.12/api/#flask.json.jsonify

majesticfish:
@dqu123 
Done

donut/modules/groups/helpers.py
```diff
@@ -40,3 +40,41 @@ def get_group_list_data(fields=None, attrs={}):
...
+    Returns:
+        result:   The fields and corresponding values of group with group_id.
+                  In the form of a dict with key:value of columnname:columnalue
+    """
+    all_returnable_fields = [
+        "group_id", "group_name", "group_desc", "type", "anyone_can_send",
+        "members_can_send", "newsgroups", "visible", "admin_control_members"
+    ]
+    default_fields = ["group_id", "group_name", "group_desc", "type"]
+    if fields == None:
```
dqu123: `if fields is None`
See https://www.python.org/dev/peps/pep-0008/#programming-recommendations
("Comparisons to singletons ...")

donut/modules/groups/helpers.py
```diff
@@ -40,3 +40,41 @@ def get_group_list_data(fields=None, attrs={}):
...
+        fields = default_fields
+    else:
+        if any(f not in all_returnable_fields for f in fields):
+            return "Invalid field"
```
dqu123: It might be better to return `None`, and then in the route return an error response (HTTP 403 Forbidden for example) if `get_group_data` returns `None`.
Or we can just check for `"Invalid field"` in the route.

dqu123 (changes requested):

donut/modules/groups/helpers.py
```diff
@@ -40,3 +40,41 @@ def get_group_list_data(fields=None, attrs={}):
...
+
+    # Build the SELECT and FROM clauses
+    s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("groups"))
+
+    # Build the WHERE clause
+    s = s.where(sqlalchemy.text("group_id = :g"))
+
+    # Execute the query
+    result = flask.g.db.execute(s, g=group_id).first()
+
```
dqu123, line 77: `result` can be `None` if there is no group with the group id.
We should return `{}` in this case. 
Currently, this causes an error when you try to `zip(fields, None)`

majesticfish: Done

dqu123 (approved):

PR #47, opened 2017-11-10
Added endpoint for listing members by group

davistardif:
### Summary
- Created endpoint for members by group (/1/groups/4/members/)

### Test Plan
- test on http://localhost:9000/1/groups/<group_id>/members/
- Also check http://localhost:9000/1/members/<user_id>
- Run `make test`

### DIFF
--- a/donut/modules/core/helpers.py
+++ b/donut/modules/core/helpers.py
@@ -45,7 +45,9 @@ def get_member_data(user_id, fields=None):
 
     # Return the row in the form of a dict (or list of dicts)
     result = [{f: t for f, t in zip(fields, res)} for res in result]
-    if len(result) == 1:
+    if len(result) == 0:
+        return {}
+    elif len(result) == 1:
         return result[0]
     else:
         return result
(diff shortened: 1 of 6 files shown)


davistardif:
Currently uses [Core Module](https://github.com/ASCIT/donut/blob/master/donut/modules/core/helpers.py) to get data about each user id. Would be more efficient to use a join instead of generating a query for each user id, but to do that this module needs access to `default_fields` from the function `get_member_data(user_id, fields=None)` in the core module.

dqu123:
@davistardif yeah it's pretty inefficient to make a database call for each user.
You can adapt `get_member_data` to take in a list of user_ids, and for the case where you want a single member, to just return the single result instead of a list.

davistardif:
@dqu123 Added you suggestion. Adding core module to test plan in case I broke anything while editing `get_member_data`

davistardif:
Ready for review

majesticfish (approved):
looks good to me!

PR #48, opened 2017-11-10
Groups positions API

calebsander:
### Summary
Adds a `GET /1/groups/<int:group_id>/positions/` API that lists all positions with their ids and names for a given group. Also resets test database when tests are run.

### Test Plan
- Go to `http://localhost:9000/1/groups/1/positions/`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,7 @@ pip-install:
 lint:
 	yapf -i -r .
 
-test:
-	make init-test-db && \
+test: init-test-db
 	python -m pytest .
 
 init-test-db:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -13,4 +13,8 @@ lint:
 	yapf -i -r .
 
 test:
+	make init-test-db && \
```
dqu123: You can make this a Makefile dependency

dqu123 (changes requested):

dqu123 (approved):

PR #49, opened 2017-11-10
Improve test coverage

dqu123:
### Summary
- Add database and transactions to client fixture
- Update Flask version so `jsonify` will handle lists
- Will potentially separate database into a separate fixture, but may not be necessary
   since most tests will need the db.

### Test Plan
- `make test` (test various core functions)


RobertEng:
lgtm

RobertEng (approved):

PR #55, opened 2017-11-17
Zmo/get group list by member endpoint

koivyr:
### Summary
- Created endpoints to get a list of groups of a member. 

### Test Plan
- Test on http://localhost:9000/1/members/<user_id>/groups/
(user_id 1 should return dev team, user_id 3 returns null)
- Run make test

### DIFF
--- a/donut/modules/core/helpers.py
+++ b/donut/modules/core/helpers.py
@@ -119,3 +119,23 @@ def get_name_and_email(user_id):
 
     result = list(flask.g.db.execute(s, {'u': user_id}))
     return (result[0][0], result[0][1])  # convert from a 2d list to a 1d list
+
+def get_group_list_of_member(user_id):
+    """
+    Queries the database and returns list of groups for a given id
+    Arguments:
+        user_id: The user_id for query.
+    Returns:
+        result: All the groups that an user_id is a part of
+    """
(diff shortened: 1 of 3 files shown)


donut/modules/core/helpers.py
```diff
@@ -119,3 +119,26 @@ def get_name_and_email(user_id):
...
+        result: All the groups that an user_id is a part of
+    """
+    s = sqlalchemy.sql.select(["group_id, group_name"]).select_from(
+        sqlalchemy.text(
+            " group_members NATURAL JOIN groups NATURAL JOIN members"))
+
+    s = s.where(sqlalchemy.text("user_id = :u"))
+    result = flask.g.db.execute(s, u=user_id)
+
+    finResult = {}
```
dqu123: Python typically uses snake case (i.e. `fin_result`).

donut/modules/core/routes.py
```diff
@@ -29,3 +29,12 @@ def get_members_list():
 def get_members(user_id):
     """GET /1/members/<int:user_id>/"""
     return json.dumps(helpers.get_member_data(user_id))
+
+
+@blueprint.route("/1/members/<int:user_id>/groups/")
+def get_group_list_of_member(user_id):
+    '''- /1/members/3/groups/ group_members table grabbing the groups
```
dqu123: You can remove the leading `-`.
Also maybe just keep `/1/members/3/groups/`, and the part that says `List all groups that a member is in`

dqu123 (changes requested):

dqu123 (approved):

PR #57, opened 2017-11-21
get_positions endpoint

kushaltirumala:
### Summary
added get_positions endpoint, which returns a list of users (user_id, group_id, pos_id, pos_name) who hold positions in a group


donut/modules/groups/helpers.py
```diff
@@ -82,3 +82,45 @@ def get_group_data(group_id, fields=None):
...
+    
+
+    fields = ["user_id", "pos_id", "group_id"]
+    s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("position_holders"))
+    result = flask.g.db.execute(s)
+    if result is None:
+        return {}
+
+    user_position_arr = []
+    for row in result:
```
calebsander: Could use a list comprehension here to map `result` to `user_position_arr`

donut/modules/groups/helpers.py
```diff
@@ -82,3 +82,45 @@ def get_group_data(group_id, fields=None):
...
+    if result is None:
+        return {}
+    
+    res = {}
+    for row in result:
+        res[row["pos_id"]] = row["pos_name"]
+    
+
+    fields = ["user_id", "pos_id", "group_id"]
+    s = sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("position_holders"))
```
calebsander: Rather than run 2 queries, you could just join the two tables `WHERE positions.pos_id = position_holders.pos_id`

donut/modules/groups/routes.py
```diff
@@ -18,6 +18,10 @@ def get_groups_list():
         fields = [f.strip() for f in flask.request.args["fields"].split(',')]
     return jsonify(helpers.get_group_list_data(fields=fields, attrs=attrs))
 
+@blueprint.route("/1/positions/")
+def get_positions():
+    return json.dumps(helpers.get_position_data())
```
calebsander: Can use `jsonify` here; takes care of setting the `Content-Type: application/json` for you

tests/modules/groups/test_groups.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+from donut.testing.fixtures import client
+from donut import app
+from donut.modules.groups.helpers import get_position_data
+from donut.modules.groups.routes import get_positions
+
+
+# Helpers
+def test_get_position_data(client):
+    assert get_position_data()[0]["user_id"] == 2
```
calebsander: Should probably test on the entirety of `get_position_data()`, not just a specific field

donut/modules/groups/routes.py
```diff
@@ -18,6 +18,10 @@ def get_groups_list():
         fields = [f.strip() for f in flask.request.args["fields"].split(',')]
     return jsonify(helpers.get_group_list_data(fields=fields, attrs=attrs))
 
+@blueprint.route("/1/positions/")
+def get_positions():
+    return jsonify(result = helpers.get_position_data())
```
RobertEng: We're using a more recent version of jsonify which doesn't require "result=". pip upgrade flask.

donut/modules/groups/helpers.py
```diff
@@ -82,3 +82,28 @@ def get_group_data(group_id, fields=None):
...
+    if fields is None:
+        fields = default_fields
+    else:
+        if any(f not in all_returnable_fields for f in fields):
+            return "Invalid field"
+
+    #s= sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("positions JOIN position_holders WHERE positions.pos_id = position_holders.pos_id"))
+    
+    
+    result = flask.g.db.execute("SELECT user_id, positions.pos_id, pos_name, positions.group_id \nFROM positions JOIN position_holders WHERE positions.pos_id = position_holders.pos_id;")
```
RobertEng: NATURAL JOIN the tables (matches the JOIN automatically on matching columns).

Also fields should be used to build up this query because otherwise you get more rows than necessary, and that would lead to incorrect results below.

RobertEng (changes requested):

donut/modules/groups/helpers.py
```diff
@@ -82,3 +82,28 @@ def get_group_data(group_id, fields=None):
...
+
+    #s= sqlalchemy.sql.select(fields).select_from(sqlalchemy.text("positions JOIN position_holders WHERE positions.pos_id = position_holders.pos_id"))
+    
+    
+    result = flask.g.db.execute("SELECT user_id, positions.pos_id, pos_name, positions.group_id \nFROM positions JOIN position_holders WHERE positions.pos_id = position_holders.pos_id;")    
+    
+    if result is None:
+        return {}
+
+    user_position_arr = [{f: t for f,t in zip(fields, row)} for row in result]
```
joshc: Here, if the fields you request are anything other than default_fields, you may return incorrect results since it would zip up all of the fields specified in the argument, since the query grabs all of the default_fields from the database.

Also, I'm not sure if you want to return user_id (since user_id is not in all_returnable_fields). Look into it.

RobertEng (dismissed):

donut/modules/groups/helpers.py
```diff
@@ -129,6 +129,35 @@ def get_group_data(group_id, fields=None):
...
+
+    if fields is None:
+        fields = default_fields
+    else:
+        if any(f not in all_returnable_fields for f in fields):
+            return "Invalid field"
+
+    field_string = "positions.pos_id"
+    for field in fields:
+        if (field == "group_id"):
```
RobertEng: no need for parenthesis

RobertEng (approved):
lgtm

PR #59, opened 2018-01-09
Updating Style for Website Redesign

Thought-Weaver:
The style.css and layout.html are the two important changes. There were some accidental changes to the helpers.py and routes.py in groups that I manually reverted, so just ignore those.


RobertEng (dismissed):
Make sure the endpoints you accidentally changed don't end up affecting anything.

lgtm

donut/modules/groups/routes.py
```diff
@@ -35,9 +35,3 @@ def get_groups(group_id):
 def get_group_members(group_id):
     """GET /1/groups/<int:group_id>/"""
     return jsonify(helpers.get_members_by_group(group_id))
-
-
-@blueprint.route("/1/positions/<int:pos_id>/")
```
dqu123, line 40: @RobertEng did you intend for this to be removed? I think it's breaking the test build right now.

Thought-Weaver: I can take a look at them tomorrow. I'll go back to one of the branches and see if I can revert these endpoints to what they're supposed to be.

PR #61, opened 2018-01-12
Added some endpoints to help positions list

majesticfish:
### Summary
- Modified the members/1/groups endpoint to become a list and include information on if the member is an admin in the particular group
- Added positions/1/ endpoint to gather list of all people associated with a position
### Test Plan
- Added test cases


RobertEng (approved):
lgtm

PR #62, opened 2018-01-30
Py mysql conversion

RobertEng:
### Summary
- start to pymysql conversion. Wrote up the connection junk.
- Converted members endpoints

### Test Plan
- members endpoints


joshc (approved):
lgtm

PR #65, opened 2018-02-09
Fixed footer being on top of things

calebsander:
### Summary
- Footer is now actually a footer
- Fixed `<div>` tags with no matching close tag

### Test Plan
- Open pages with footers on them


RobertEng (approved):
lgtm

PR #68, opened 2018-03-13
footer fix

RobertEng:
### Summary
- The footer looks attached to the bottom now.

### Test Plan
- Look at it.

@joshc @mhuynh5757 @Thought-Weaver


RobertEng:
Does not work on all laptops for some reason. Moving this to the updated style branch

Issue #10, opened 2015-09-21
Things to think about when making directory search


allisonkong:
Just a list of things that came to mind (may update this later):
- Prefrosh emails should be hidden during rotation
- Gender options should **not** be limited to male/female (option to fill in their own string?)
  - Use an enum for gender (`Gender.Male`, `Gender.Female`, `Gender.Other`) so the gender is stored as an int in the database, and then have a second (nullable) column for gender_string so you can store a custom string if `Gender.Other` is used. 
- Search by zip code is kinda creepy, honestly (and also is pretty broken)
- A feature that would be cool is search-as-you-type (like google search)

epelz:
As a note, we got rid of search by zip code (for non-admins). We replaced it with displaying hometown on the directory result, but you can't search by it.

allisonkong:
Ah, didn't realize that was admin only.

PR #69, opened 2018-03-13
Directory search

calebsander:
### Summary
- Directory search, results, and view user page
- Importing directory data from legacy server (except for groups & positions)
- Database changes:
  - Using separate buildings table
  - Allowing non-numeric room numbers
  - Allowing longer option names

### Test Plan
- Search some people on directory
- Look at your own page by clicking username in top right
- Try changing your pro pic
- Try setting custom gender and nickname


donut/constants.py
```diff
@@ -15,3 +15,9 @@
 # Length of time before recovery key expires, in minutes.
 PWD_RESET_KEY_EXPIRATION = 1 * 24 * 60
 CREATE_ACCOUNT_KEY_LENGTH = 32
+"""Value of members.gender if member's gender is unknown"""
```
dqu123: Use an enum
https://docs.python.org/3/library/enum.html
Apparently you can have multiple aliases to the same value, and you could have a custom constructor
to put all the conversion logic in one place.

dqu123 (changes requested):
Nice work! I tried it out and it's starting to really take shape. Looking forward to a MVP sometime soon :)

donut/modules/directory_search/templates/view_user.html
```diff
@@ -0,0 +1,159 @@
...
+				<ul class='list-group list-group-root'>
+					{% if user['username'] %}
+						<li class='list-group-item'>
+							<strong>Username</strong>:
+							{{ user['username'] }}
+						</li>
+					{% endif %}
+					<li class='list-group-item'>
+						<strong>UID</strong>:
+						{{ user['uid'] }}
```
dqu123: Should check `is_me`. UID is personally identifiable information (PII).

donut/modules/directory_search/templates/view_user.html
```diff
@@ -0,0 +1,159 @@
...
+					</li>
+					{% if user['birthday'] %}
```
dqu123: Should check `is_me`. Birthday is PII. For example it is used to sign up and has been used for elections as well.

donut/modules/directory_search/templates/view_user.html
```diff
@@ -0,0 +1,159 @@
...
+							{{ user['gender_string'] }}
+						</li>
+					{% endif %}
+					{% if user['email'] %}
+						<li class='list-group-item'>
+							<strong>Email</strong>:
+							<a href='mailto:{{ user["email"] }}'>{{ user['email'] }}</a>
+						</li>
+					{% endif %}
+					{% if 'phone_string' in user %}
```
dqu123: Again, PII, should consider a privacy / permissions setting for this. Currently RAs and some ASCIT positions can see it.

dqu123 (changes requested):
@calebsander  @kushaltirumala @RobertEng @joshc 
@mhuynh5757 We need to be careful about privacy and sensitive data like UIDs, birthdays, phone numbers, etc. #10 has some relevant points (I mentioned the gender enum in the previous review).

donut/modules/directory_search/templates/view_user.html
calebsander: Yeah, I was waiting to implement hiding information until a permissioning system was in place. Hiding personal information from other people is simple enough, but I'd imagine we would want admin and RA privileges to override it. The `get_permissions` function in `auth_utils` seems like it should support this, but it references tables that don't exist.

calebsander:
Did the gender `Enum` thing and added a basic framework for hiding PII; should work once permissions system is implemented, but for now no one will be treated as an admin

donut/modules/directory_search/helpers.py
```diff
@@ -0,0 +1,244 @@
...
+            user['gender_string'] = 'Female'
+        phone = user['phone']
+        if phone:
+            if len(phone) == 10 and all(['0' <= d <= '9' for d in phone]):
+                user['phone_string'] = '(' + phone[:3] + ') '
+                user['phone_string'] += phone[3:6] + '-' + phone[6:]
+            else:  #what sort of phone number is that
+                user['phone_string'] = phone
+        state = user['state']
+        if state:
```
RobertEng: For international students, it looks like it prints out an empty string for state.

donut/modules/directory_search/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+        return cursor.fetchone()['preferred_name'] or ''
+
+
+def set_preferred_name(user_id, name):
+    query = 'UPDATE members SET preferred_name = %s WHERE user_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [name, user_id])
+
+
+def get_gender(user_id):
```
RobertEng: Move these helper functions and route endpoints over to core.
Some functions which query member are a rehash of the get_member_data helper in core and should use that instead.

donut/modules/directory_search/routes.py
```diff
@@ -0,0 +1,149 @@
...
+    return flask.render_template(
+        'directory_search.html',
+        houses=helpers.get_houses(),
+        options=helpers.get_options(),
+        residences=helpers.get_residences(),
+        grad_years=helpers.get_grad_years(),
+        states=helpers.get_states())
+
+
+@blueprint.route('/1/users/me')
```
RobertEng: Move /user/ endpoints to core module

sql/donut.sql
```diff
@@ -54,20 +65,21 @@ CREATE VIEW members_full_name AS (
 -- Major Option Table
 CREATE TABLE options (
     option_id     INT         NOT NULL AUTO_INCREMENT,
-    option_name   VARCHAR(16) NOT NULL,
-    PRIMARY KEY (option_id)
+    option_name   VARCHAR(50) NOT NULL,
```
RobertEng, line 68: 64

RobertEng (changes requested):

RobertEng:
Also add a page on the wiki with quick details on the directory

calebsander:
Okay, fixed everything that was mentioned except for the 64-wide `VARCHAR` thing, since apparently anything any value up to 255 results in the same representation on disk

RobertEng:
lgtm

RobertEng (approved):

PR #73, opened 2018-04-24
Implemented changes to positions and group position memberships

majesticfish:
### Summary
- Deleted group_members table
- Changed queries of group_members table to quires of position_holders NATURAL JOIN positions
   - This adheres to the paradigm that all members of groups have an associated position
- Created a position relation table
### Test Plan
-

### DIFF
--- a/donut/modules/core/helpers.py
+++ b/donut/modules/core/helpers.py
@@ -128,8 +128,10 @@ def get_group_list_of_member(user_id):
     Returns:
         result: All the groups that an user_id is a part of
     """
-    query = """SELECT group_id, group_name, control
-    FROM positions NATURAL JOIN groups NATURAL JOIN position_holders
+    query = """SELECT DISTINCT group_id, group_name, control
+    FROM groups NATURAL JOIN positions p LEFT JOIN position_relations pr
+    ON p.pos_id=pr.pos_id_to INNER JOIN position_holders ph ON
+    ph.pos_id=p.pos_id OR pr.pos_id_from=ph.pos_id
     WHERE user_id = %s"""
 
     with flask.g.pymysql_db.cursor() as cursor:
--- a/donut/modules/core/helpers.py
+++ b/donut/modules/core/helpers.py
@@ -129,7 +129,7 @@ def get_group_list_of_member(user_id):
         result: All the groups that an user_id is a part of
     """
     query = """SELECT group_id, group_name, control
-    FROM group_members NATURAL JOIN groups NATURAL JOIN members
+    FROM positions NATURAL JOIN groups NATURAL JOIN position_holders
     WHERE user_id = %s"""
 
     with flask.g.pymysql_db.cursor() as cursor:
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -138,7 +138,22 @@ def get_position_data(fields=None):
 
 
 def get_members_by_group(group_id):
-    query = "SELECT user_id FROM positions NATURAL JOIN position_holders "
+    '''
+    Queries the database and returns a list of all users associated with
+    a particular group either because a) They hold a position in the group
+    or b) They hold a position linked to another position in the group
+
+    Arguments:
+        group_id: id of group in question
+    
+    Returns:
+        List where each element is a JSON reprenting the data of each
+        person
+    '''
+    query = "SELECT DISTINCT user_id FROM positions p LEFT JOIN "
+    query += "position_relations pr ON p.pos_id=pr.pos_id_to "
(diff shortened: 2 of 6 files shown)


sql/donut.sql
```diff
@@ -173,32 +179,11 @@ CREATE VIEW group_house_membership AS (
...
-                                       -- can send emails to group
-    control     BOOLEAN DEFAULT FALSE, -- Toggles whether or not this position
-                                       -- has admin control over group
-    receive     BOOLEAN DEFAULT TRUE,  -- Toggles if this position receives
-                                       -- emails from this group
-    PRIMARY KEY (pos_id, group_id),
-    FOREIGN KEY (pos_id) REFERENCES positions(pos_id),
-    FOREIGN KEY (group_id) REFERENCES groups(group_id)
-);
+CREATE TABLE position_relations (
```
RobertEng: Are there any tests for this table?

donut/modules/core/helpers.py
```diff
@@ -129,7 +129,7 @@ def get_group_list_of_member(user_id):
         result: All the groups that an user_id is a part of
     """
     query = """SELECT group_id, group_name, control
-    FROM group_members NATURAL JOIN groups NATURAL JOIN members
+    FROM positions NATURAL JOIN groups NATURAL JOIN position_holders
```
calebsander: Doesn't this need to take `positions_relations` into account?

donut/modules/groups/helpers.py
```diff
@@ -138,7 +138,7 @@ def get_position_data(fields=None):
 
 
 def get_members_by_group(group_id):
-    query = "SELECT user_id FROM group_members "
+    query = "SELECT user_id FROM positions NATURAL JOIN position_holders "
```
calebsander: Doesn't this need to take `positions_relations` into account?

donut/modules/core/helpers.py
```diff
@@ -129,7 +129,7 @@ def get_group_list_of_member(user_id):
         result: All the groups that an user_id is a part of
     """
```
RobertEng, line 130: Should this function be called get_member_groups

sql/donut.sql
```diff
@@ -173,32 +179,11 @@ CREATE VIEW group_house_membership AS (
...
+    relation_id   INT  NOT NULL AUTO_INCREMENT,
```
calebsander: Could just use `(pos_id_from, pos_id_to)` as your PK

calebsander (changes requested):
Generally looks good, although I think you aren't making use of `positions_relations`. Do the tests all still work? Also, I think we should remove the `group_id` column from `position_holders` since it should be specified by the referenced position.

sql/donut.sql
```diff
@@ -173,32 +179,11 @@ CREATE VIEW group_house_membership AS (
...
+    pos_id_from   INT  NOT NULL,
+    pos_id_to     INT  NOT NULL,
```
RobertEng: It looks like this table should be used in some helper functions but is not. Add a comment for what this table is.

sql/donut.sql
```diff
@@ -139,6 +138,13 @@ CREATE TABLE positions (
     group_id INT         NOT NULL,
     pos_id   INT         NOT NULL AUTO_INCREMENT,
     pos_name VARCHAR(32) NOT NULL,
+    send        BOOLEAN DEFAULT FALSE, -- Toggles whether or not this position
```
RobertEng, line 141: send, control, and receive have their defaults defined here, but also need to be changed on a person by person basis.

RobertEng (changes requested):

sql/donut.sql
majesticfish: Sorry Whats the problem with that? The default is just the value that is set when theres nothing there right?

sql/donut.sql
majesticfish: no I'll write them

donut/modules/core/helpers.py
majesticfish: yup

joshc (approved):
lgtm - further work needed on newsgroups part

sql/test_data.sql
```diff
@@ -47,6 +47,9 @@ INSERT INTO members(
     1,
     '203'
 );
+INSERT INTO members(uid, last_name, first_name, email, phone) VALUES
+    ('2045251', 'Yu', 'Sean', 'ssyu@caltech.edu', NULL);
```
calebsander, line 51: `phone` doesn't have to be specified, but minor point

calebsander (approved):

PR #75, opened 2018-05-10
Voting

calebsander:
### Summary
- Support for surveys/elections:
  - Can create and edit surveys, take them, and view the responses once they close
  - Currently supported question types: dropdown (select one from a list), checkboxes (select a subset of a list), elected position (allows ordering a list, submitting write-ins, and NO; results of ranked pairs are shown), short text (input), and long text (textarea)
  - Surveys can be restricted to members of a specific group
  - A survey can be configured not to appear in the list of active surveys so that a link is required to take it
  - Can also require survey-takers to authenticate with their UID and birthday
  - Survey results are available to the creator once the survey closes, but must be explicitly released by the creator so others can see them
- Refactored the directory search name results display so it is easy to add it to other pages
- Also figured out why tests were taking 2 minutes to load and fixed it

### Test Plan
- Try creating some surveys, editing their questions and settings
- Make sure questions editor has all the features we might want to provide
- Have a bunch of users take a test survey so the results can be verified


donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+
+def get_question_types():
+    global question_types
+    if not question_types:
+        query = 'SELECT type_id, type_name FROM survey_question_types'
+        with flask.g.pymysql_db.cursor() as cursor:
+            cursor.execute(query)
+            question_types = {
+                question_type['type_name']: question_type['type_id']
+                for question_type in cursor.fetchall()
```
joshc, line 35: dictionary comprehension :O

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+            }
+    return question_types
+
+
+def allowed_to_take(user_id):
+    return lambda survey: survey['group_id'] is None or is_user_in_group(user_id, survey['group_id'])
+
+
+def get_public_surveys(user_id):
```
joshc: probably more appropriately named "get_visible_surveys" or "get_accessible_surveys" or something like that.

donut/modules/groups/helpers.py
```diff
@@ -182,7 +182,7 @@ def get_members_by_group(group_id):
 
     Arguments:
         group_id: id of group in question
-    
+
     Returns:
         List where each element is a JSON reprenting the data of each
```
joshc: representing

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+import json
+import flask
+from donut.auth_utils import get_user_id
+from donut.misc_utils import generate_random_string
+from donut.modules.groups.helpers import get_group_list_data, is_user_in_group
+from donut.validation_utils import (validate_date, validate_exists,
+                                    validate_in, validate_int)
+from .ranked_pairs import winners
+
+ACCESS_KEY_LENGTH = 64
```
joshc, line 13: this is long :( what is the purpose of the access key? why not just index it. see overall comments

donut/modules/voting/templates/edit.html
```diff
@@ -0,0 +1,273 @@
+{% extends "layout.html" %}
```
joshc, line 1: personally like the two spaces indentation better than the tab indentation (makes it harder to read) in other html files

donut/static/css/style.css
```diff
@@ -6,6 +6,17 @@
   box-sizing: inherit;
```
joshc, line 6: won't comment on css

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+
+def get_question_type(question_id):
+    query = 'SELECT type_id FROM survey_questions WHERE question_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [question_id])
+        return cursor.fetchone()['type_id']
+
+
+def get_choice(question_id, value):
+    query = 'SELECT choice_id FROM survey_question_choices WHERE question_id = %s AND choice = %s'
```
joshc: why would you need this? choice ids are all unique by definition, so you could associate each visible choice with a choice id instead of needing to extract it like this. also, since you don't restrict duplicate values for a single question, it's ambiguous which choice a question_id + value combo is referring to which may or may not be relevant.

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+    insert_choice_query = """
+        INSERT INTO survey_question_choices (question_id, choice)
+        VALUES (%s, %s)
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(delete_query, [survey_id])
+        for order, question in enumerate(questions):
+            cursor.execute(insert_question_query, [
+                survey_id, question['title'], question['description'].strip()
```
joshc: should also strip title?

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+    if user_id != survey['creator']:
+        return 'You are not the creator of this survey'
+    if not allow_after_close and survey['end_time'] < datetime.now():
+        return 'Cannot modify a survey after it has closed'
+
+
+def set_responses(question_ids, responses):
+    user_id = get_user_id(flask.session['username'])
+    query = """
+        INSERT INTO survey_responses(question_id, user_id, response)
```
joshc, line 323: commenting a bit on the table structure. i believe the legacy site has `user_id`, `choice_id`, `response`, with `choice_id` being the choice selected and `response` being text in the case of a text entry. sure, you can represent all choices by text in the responses column, but it probably makes joins a little dirtier if you ever need to join (instead of well-defined numbers, you'd be joining on text). up to your discretion about whether to use `choice_id` here or not.

donut/modules/voting/routes.py
```diff
@@ -0,0 +1,264 @@
...
+@blueprint.route('/1/surveys/<access_key>/questions', methods=['GET'])
+def edit_questions(access_key):
+    survey = helpers.get_survey_data(access_key)
+    restrict_message = helpers.restrict_edit_access(survey, False)
+    if restrict_message:
+        flask.flash(restrict_message)
+        return list_surveys()
+
+    survey_id = survey['survey_id']
+    if helpers.some_responses_for_survey(survey_id):
```
joshc: seems useless unless you prompt the user if they want to continue. in fact, i don't think the warning message below ever shows up, in my brief test of this. two ways i see to handle this: prompt and return if they don't want to erase all responses, or give a warning at survey creation saying that they won't be able to change the settings on the survey if the survey is currently active.

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+        questions = cursor.fetchall()
+        for question in questions:
+            if question['choices']:
+                cursor.execute(choices_query, [question['question_id']])
+                question['choices'] = {
+                    choice['choice_id']: choice['choice']
+                    for choice in cursor.fetchall()
+                }
+            else:
+                del question['choices']
```
joshc: probably better ways to handle what this is doing than with a `del`

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+            cursor.execute(responses_query, question['question_id'])
+            responses = [
+                json.loads(res['response']) for res in cursor.fetchall()
+            ]
+            question['responses'] = responses
+            question_type = question['type']
+            if question_type in count_types:
+                question['results'] = Counter(responses).most_common()
+            elif question_type == question_types['Checkboxes']:
+                question['results'] = Counter(chain(*responses)).most_common()
```
joshc: what is this party.

donut/modules/voting/ranked_pairs.py
```diff
@@ -0,0 +1,50 @@
+from itertools import product
+
+WINNERS_TO_LIST = 3
+
+
+def winners(responses):
```
joshc, line 7: very nice. as far as i can tell seems correct, though algorithms were never my strong point haha. make sure you include everything they talk about in the bylaws. for one, and i'm unsure about this, is the ranking you define here mathematically equivalent to the ranking specified in the bylaws (with respect to picking multiple winners) where you choose the one winner, remove all votes relevant to that winner, and rerun the algorithm to generate the next winner? i am not sure if it is.
Also, it is worth mentioning that some positions require more than one winner, in which case we shouldn't display the results as rankings but rather as winners, or at least 2 first places. lol.

donut/modules/voting/routes.py
```diff
@@ -0,0 +1,264 @@
...
+        **survey,
+        is_owner=is_owner,
+        NO=helpers.NO,
+        question_types=helpers.get_question_types(),
+        questions_json=questions_json)
+
+
+@blueprint.route('/1/surveys/make')
+def make_survey_form():
+    tomorrow = date.today() + timedelta(days=1)
```
joshc: probably unnecessary to define defaults for these, since they're required for every survey and whoever is making the survey will enter them in anyways. so unless they're necessary to make the code work (i.e. variables can't be null) we should force whoever is making the survey to make sure they enter the dates in manually to make sure they are entering it correctly.

donut/modules/voting/routes.py
```diff
@@ -0,0 +1,264 @@
...
+            choice = helpers.get_choice(question_id, value)
+            if choice is None: return error('Invalid choice for dropdown')
+            response_json = choice['choice_id']
+        elif type_id == question_types['Elected position']:
+            if type(value) != list:
+                return error('Invalid response to elected position')
+            response_json = []
+            for order_value in value:
+                if type(order_value) == str:
+                    if order_value == helpers.NO: response_json.append(None)
```
joshc: why append None if choice is NO? just asking haha

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+                cursor.execute(choices_query, [question['question_id']])
+                question['choices'] = {
+                    choice['choice_id']: choice['choice']
+                    for choice in cursor.fetchall()
+                }
+            else:
+                del question['choices']
+
+            def resolve_name(vote):
+                if vote is None: return NO
```
joshc: seems doubtful that we would need to represent a "NO" vote as None, why can't it just be a "NO" string?

donut/modules/voting/helpers.py
```diff
@@ -0,0 +1,395 @@
...
+        end_time <= NOW() as closed, end_time
+        FROM surveys WHERE creator = %s
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [user_id])
+        return cursor.fetchall()
+
+
+def delete_survey(survey_id):
+    query = 'DELETE FROM surveys WHERE survey_id = %s'
```
joshc: if possible please also delete associated `survey_responses` and `survey_question_choices` since there's literally no reason they should be in the database if the survey has been deleted.

sql/voting.sql
```diff
@@ -0,0 +1,73 @@
...
+    access_key     CHAR(64)      NOT NULL, -- hex string
+    group_id       INT,                    -- restrict to this group if non-NULL
+    auth           BOOLEAN       NOT NULL, -- require authentication (UID and birthday)
+    public         BOOLEAN       NOT NULL, -- whether survey should appear in the list of active surveys
+    results_shown  BOOLEAN       DEFAULT FALSE, -- whether survey should appear in the list of results
+    creator        INT           NOT NULL, -- user id of survey creator
+
+    PRIMARY KEY(survey_id),
+    FOREIGN KEY(group_id) REFERENCES groups(group_id) ON DELETE SET NULL,
+    FOREIGN KEY(creator) REFERENCES members(user_id) ON DELETE CASCADE
```
joshc: don't want this. if a user decides to delete their account or something, we will want to keep the surveys they started as well.

sql/voting.sql
```diff
@@ -0,0 +1,73 @@
...
+    choice       VARCHAR(50)  NOT NULL,
+
+    PRIMARY KEY(choice_id),
+    FOREIGN KEY(question_id) REFERENCES survey_questions(question_id) ON DELETE CASCADE
+);
+
+CREATE TABLE survey_responses (
+    question_id  INT  NOT NULL,
+    user_id      INT,
+    response     TEXT, -- JSON
```
joshc: again, consider doing `user_id`, `choice_id`, `response`. looking at the comments you have, dropdown + checkbox + most of elected position use choice_id primarily.

donut/modules/voting/helpers.py
joshc: jk you did it below with cascades

sql/voting.sql
```diff
@@ -0,0 +1,73 @@
...
+    question_id  INT           NOT NULL AUTO_INCREMENT,
+    survey_id    INT           NOT NULL,
+    title        VARCHAR(100)  NOT NULL,
+    description  TEXT,
+    list_order   INT           NOT NULL, -- position in survey, zero-indexed
+    type_id      INT           NOT NULL,
+
+    PRIMARY KEY(question_id),
+    FOREIGN KEY(survey_id) REFERENCES surveys(survey_id) ON DELETE CASCADE,
+    FOREIGN KEY(type_id) REFERENCES survey_question_types(type_id)
```
joshc, line 43: no `ON DELETE CASCADE` here? hope types are never deleted ;)

sql/voting.sql
```diff
@@ -0,0 +1,73 @@
...
+);
+
+CREATE TABLE survey_question_choices (
+    choice_id    INT          NOT NULL AUTO_INCREMENT,
+    question_id  INT          NOT NULL,
+    choice       VARCHAR(50)  NOT NULL,
```
joshc: seems a tad short. consider long-ass question choices, like which of the following would you do in this scenario? I would `pick up my knife flip the table and exceed 50 chars`, or i would `they defenestrate, and then abbreviate people life.` etc.

sql/voting.sql
```diff
@@ -0,0 +1,73 @@
...
+    question_id  INT  NOT NULL,
+    user_id      INT,
+    response     TEXT, -- JSON
+                       -- For dropdown:         choice_id (number)
+                       -- For checkbox:         [choice_id (number)]
+                       -- For short/long text:  text (string)
+                       -- For elected position: [choice_id (number) | user_id (negative number) | NO (null)]
+    PRIMARY KEY(question_id, user_id),
+    FOREIGN KEY(question_id) REFERENCES survey_questions(question_id) ON DELETE CASCADE,
+    FOREIGN KEY(user_id) REFERENCES members(user_id) ON DELETE CASCADE
```
joshc: unsure if we want this. most likely not, i think.

joshc (changes requested):
after you've read through the review / fixed things that need to be fixed if any, i'll try to make a clone of one of this year's election on your website and see how it fares through some poking and prodding. also some of the comments i made were just making sure that you handled it lol.

random thoughts i had while reviewing/testing code:
- uneven indenting sometimes with tabs <-> spaces rip.
- add close survey button for users
- instead of showing "already completed", show them what votes they cast (and maybe allow them to change votes, though if that's too complicated logically don't do it)
- modifying a survey's end time to before now gives the following flask.flash message: "Cannot modify a survey after it has closed"
- should not automatically publish results to website once election has ended according to bylaws (24-hour "grace" period for concerns to be raised). i think you're handling it already but just making sure
- require election candidates to be people from directory too, not just write-ins (revcomm chair occasionally fucks up entering in names, which would cause them to be counted as separate candidates). consider adding support for lists of directory people as candidates (examples being tech editors, senior class presidents)
- eventually once permissions is implemented will have to restrict survey creation privilege to users (or groups)
- what is the purpose of the access key? why not just index it. permission to take survey should be be based on user id, not just on whether they're able to see it or not (or else people outside of the intended group would be able to take the survey)
- add button to view survey from survey edit page (or automatically redirect once save is clicked)
- style is bad on the survey editing/viewing page
- related to the point above, dragging to reorder the candidates looks weird
- display the default choice of NO in the adding election questions page so that whoever is making the survey knows that they don't have to add it. bonus: support elections with no NO choice (ha)
- unsure about displaying raw voting results since may take a while to load / make the page very long. perhaps just do what we have now and display some summary results, whatever those are (right now i think it's frequency of votes for each choice)
- is it possible to require an answer to a question in order to submit a vote? for example, the first question in surveys is a Yes No question "do you understand how to vote" which is currently not enforced since you can't require an answer (i tried to enforce it by only putting Yes, but people just ignore the question :( )
- also consider whether it's necessary to load legacy survey data into this system (maybe just the responses?)
- bonus: randomize the order on election ballots for fairness
- bonus: legacy system has ability to create copies of surveys. seems unnecessary anymore since your UI is much better, just mentioning here.

donut/modules/groups/helpers.py
calebsander: Yeah, not my code, but I can change it

donut/modules/voting/helpers.py
calebsander: The access key is to keep surveys private, so you can create a survey that only people with the link (which is `/<access_key>/take`) can fill out. I didn't want it to be brute-forceable but I guess 32 would probably suffice.

sql/voting.sql
calebsander: Checkboxes store arrays of `choice_id`s; elected position stores arrays of `user_id`s and `choice_id`s. I wanted each user to have a single answer to each question, so it is hard to represent the arrays without having a variable-length field. Also wanted to maximize support for adding new question types in the future.

donut/modules/voting/helpers.py
joshc: yeah, i also mention in the overall comments that you should be verifying viewing/taking/submitting the survey based on user_id anyways, which i think you do? in which case it doesn't matter if they have the link or not, they wouldn't be able to fill it out.

sql/voting.sql
joshc: makes sense and sounds good.

donut/modules/voting/helpers.py
calebsander: I think you need some sort of lookup mechanism to make sure that people aren't submitting invalid responses. For example, submitting a `choice_id` which corresponds to a different question. But yeah, I agree it is probably nicer to just submit `choice_id`s

donut/modules/voting/helpers.py
calebsander: Yeah that's only in there because I was checking `if 'choices' in question` in the results rendering, but I guess I could easily change that to `if question['choices']`

donut/modules/voting/helpers.py
calebsander: The format of the elected positions responses is arrays of choice_ids and user_ids; I didn't want to allow arbitrary strings, so I just allowed null. I imagine it makes it slightly faster for the ranked pairs algorithm because it is faster to check None for equality than checking strings for equality.

donut/modules/voting/ranked_pairs.py
calebsander: I'm pretty sure it is equivalent. Wikipedia says that you can repeatedly remove the winner and re-run the algorithm, but it also shows that you can just build the optimal DAG once and then a topological sort on it gives you the ranking.

calebsander: Doesn't seem so critical to me whether it says "1. Candidate A; 2. Candidate B; 3. Candidate C" or "Top 3: A, B, and C". I think it is unnecessary extra logic to store how many winners each elected position question has.

donut/modules/voting/routes.py
calebsander: I'm pretty sure it does show up, but it shows up whenever you open the page to edit the questions, not just when you try to save them. Probably wouldn't be too hard to have a prompt instead.

donut/modules/voting/routes.py
calebsander: Just because I am not allowing arbitrary strings in the responses

sql/voting.sql
calebsander: Yeah I think it doesn't really make sense to allow for deleting question types

sql/voting.sql
calebsander: Yeah I guess you could put more information in the question description, but we could definitely expand this to 255 characters or something

donut/modules/voting/helpers.py
calebsander: Yes, you can also restrict by group which checks the `user_id`. I was mostly emulating Google Forms' functionality, not sure whether it would be useful to send a link out to a survey to a group of people that doesn't exist in the database.

calebsander:
Alright - I think I addressed all of Joosh's concerns. I thought it would kinda hacky to only allow people as elected position options, in case we find out later that this is too restrictive. I also haven't added a "do you understand how to vote" question yet; wasn't sure how critical that is.

joshc:
I think it would be good to be slightly restrictive. Elections should be intended only for candidates who are humans (other types of elections can probably be done with Google forms / other survey types). One of the problems in the old site was that the RevComm chair would screw up in making the election and slightly misspell the name of the candidate, which meant in computing the results it'll count it as two candidates which is obviously not correct. Just keep that in mind.

joshc:
image
"If a candidate is ranked multiple times, the best ranking shall be kept, and all others shall be removed." Shouldn't this mean that the result should be Cleb, then NO, then Josh?

Also, do you have any ideas for supporting equal rankings for candidates, as indicated by the bylaws? I think that most of the infrastructure is solid, just need to make sure that some minor things are taken care of. Generally the description of the survey has a link to the candidacy statements and also the revcomm email, which we may want to add support for along with line breaks, though it's not necessary (legacy does not).

calebsander:
Alright, I fixed it so only the highest ranking of each candidate is kept. I think it is too complicated to implement orderings that allow equality; I also can't really see how one could be completely indifferent between multiple candidates. I made the descriptions also format line breaks and links on the take, edit, and results pages.

joshc:
ok sounds good to me. i think the reasoning behind the equality is mainly if you heavily disfavor a candidate against multiple others (so somewhat relative). lgtm, going to approve changes so you can merge. there are some style things that i assume are fixed by logan's style updates. also, it would be nice to have the data from the old elections on the new site but not necessary/possible depending on the differences in formats, so some sort of upload script.

joshc (approved):

PR #76, opened 2018-05-15
hotfix pymysql

RobertEng:
### Summary
- Hotfix for the rest of sqlalchemy to pymysql conversion. Missed a few instances and was crashing account creation.

### Test Plan
- There were no tests for account creation other than manually.

@joshc

### DIFF
--- a/donut/__init__.py
+++ b/donut/__init__.py
@@ -74,7 +74,7 @@ def before_request():
     if 'DB_URI' in app.config:
         engine = sqlalchemy.create_engine(
             app.config['DB_URI'], convert_unicode=True)
-        flask.g.db = engine.connect() 
+        flask.g.db = engine.connect()
     """Logic executed before request is processed."""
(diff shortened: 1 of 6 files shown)


calebsander (dismissed):
lgtm

calebsander:
Except `make lint` is failing

joshc (commented):
Seems like some things are failing.
image

joshc (changes requested):
Ran `make lint` which yielded the errors from `make test`. Please fix those errors. Seems like there still needs to be some clean up done with regards to how transactions are differently represented in pymysql vs sqlalchemy.

calebsander:
I think we should probably leave `sqlalchemy` in as a requirement for now so that the marketplace tests don't fail. Once Thomas merges in the new marketplace we should be able to get rid of it.

RobertEng:
Yea, that's what i was thinking

joshc (approved):

joshc:
lgtm

PR #77, opened 2018-06-05
Changed Methods to Take into Account Position Relations

majesticfish:
### Summary
Changed get_position_data and get_position_holders methods to account for the case when someone holds a position through a position relation and not necessarily directly

### Test Plan
Also altered test cases


davistardif (approved):
lgtm

PR #78, opened 2018-06-05
Dtardif/permissions

davistardif:
### Summary
- Added some simple SQL and python to handle permissions. 

### Test Plan
- make test


davistardif:
Sorry if the commit history is confusing. I accidentally started it from the wrong branch and had to rebase to master.

sql/permissions.sql
```diff
@@ -0,0 +1,17 @@
+DROP TABLE IF EXISTS permissions;
+DROP TABLE IF EXISTS position_permissions;
+
+CREATE TABLE permissions (
+       permission_id INT NOT NULL AUTO_INCREMENT,
+       permission_name VARCHAR(255),
+PRIMARY KEY(permission_id)
```
majesticfish: Nit: the formatting appears to be off here?

sql/permissions.sql
```diff
@@ -0,0 +1,17 @@
...
+);
+
+CREATE TABLE position_permissions (
+       pos_id INT NOT NULL,
+       permission_id INT NOT NULL,
+FOREIGN KEY(pos_id) REFERENCES positions(pos_id)
```
majesticfish: See above

donut/modules/permissions/helpers.py
```diff
@@ -0,0 +1,26 @@
...
+    # get all position id's with this permission
+    query = '''SELECT pos_id FROM position_permissions WHERE permission_id = %s'''
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (permission_id))
+        result = cursor.fetchall()
+    pos_ids = [row['pos_id'] for row in result]
+    for pos_id in pos_ids:
+        holders = groups.get_position_holders(pos_id)
+        print(pos_id)
+        print(holders)
```
majesticfish: Remove print statements

sql/test_data.sql
```diff
@@ -83,8 +83,17 @@ INSERT INTO position_holders(pos_id, user_id) VALUES
...
+
 INSERT INTO position_relations(pos_id_from, pos_id_to) VALUES
     (4, 5);
 
 INSERT INTO rooms(room_id, location, title, description) VALUES
     (1, 'SAC 23', 'ASCIT Screening Room', 'A room for watching DVDs and videos');
+
+INSERT INTO permissions(permission_id, permission_name) VALUES
+    (1, 'Ruddock full member'),
+    (2, 'IHC member');
```
majesticfish: Also  minor nit: 
It'll be better for readability later if you gave these better names. Right now they sounds more like position names. Perhaps you can do like "Rudocck Website Access" or "Can view Rotation Violations" to be more clear as to what they are

majesticfish (dismissed):
Some minor things, looks good though

sql/permissions.sql
```diff
@@ -0,0 +1,17 @@
+DROP TABLE IF EXISTS permissions;
+DROP TABLE IF EXISTS position_permissions;
```
calebsander, line 1: These `DROP`s need to be switched; you break a foreign key constraint if you drop `permissions` while `position_permissions` still exists

donut/modules/permissions/helpers.py
```diff
@@ -0,0 +1,26 @@
...
+        holders = [row['user_id'] for row in holders]
+        if user_id in holders:
```
calebsander: Could just check `if {'user_id': user_id} in holders`

sql/test_data.sql
```diff
@@ -83,8 +83,17 @@ INSERT INTO position_holders(pos_id, user_id) VALUES
...
     (5, 4);
 
+
 INSERT INTO position_relations(pos_id_from, pos_id_to) VALUES
     (4, 5);
 
 INSERT INTO rooms(room_id, location, title, description) VALUES
     (1, 'SAC 23', 'ASCIT Screening Room', 'A room for watching DVDs and videos');
+
+INSERT INTO permissions(permission_id, permission_name) VALUES
```
calebsander: Shouldn't have to specify `permission_id` here, since it auto-increments

donut/modules/permissions/helpers.py
```diff
@@ -0,0 +1,26 @@
...
+    Returns True if [user_id] holds a position that directly 
+    or indirectly (through a position relation) grants 
+    them [permission_id]. Otherwise returns False.
+    '''
+    if not (isinstance(user_id, int) and isinstance(permission_id, int)):
+        return False
+    # get all position id's with this permission
+    query = '''SELECT pos_id FROM position_permissions WHERE permission_id = %s'''
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (permission_id))
```
calebsander: Parens unnecessary

calebsander (changes requested):

sql/test_data.sql
davistardif: Just to make it clear when comparing the tests to the test data

donut/modules/permissions/helpers.py
davistardif: Not sure why, but that didn't work for me

calebsander (dismissed):

donut/modules/permissions/helpers.py
```diff
@@ -0,0 +1,24 @@
...
+def has_permission(user_id, permission_id):
+    ''' 
+    Returns True if [user_id] holds a position that directly 
+    or indirectly (through a position relation) grants 
+    them [permission_id]. Otherwise returns False.
+    '''
+    if not (isinstance(user_id, int) and isinstance(permission_id, int)):
+        return False
+    # get all position id's with this permission
+    query = '''SELECT pos_id FROM position_permissions WHERE permission_id = %s'''
```
RobertEng: Instead of using a mix of sql and python, try converting this entire method into one query which grabs user_ids

RobertEng (changes requested):

donut/modules/permissions/helpers.py
calebsander: I think it's hard because there is a rather complex query needed to find direct and indirect position holders in a group

RobertEng: ... unless the groups.get_position_holders method is crazy and needs to be done separately

donut/modules/permissions/helpers.py
```diff
@@ -0,0 +1,24 @@
+import flask
+from donut.modules.groups import helpers as groups
+
+
+def has_permission(user_id, permission_id):
```
RobertEng: How will modules know which permission_id to call on? It should be some descriptor of the permission instead, either has_permissions(user_id, permission_name) or an ENUM of permissions stored in a file.

RobertEng (dismissed):

donut/modules/permissions/helpers.py
davistardif: I'm with Caleb on this -- could do this as one query but I would be duplicating what's already written in groups as a subquery

donut/modules/permissions/helpers.py
davistardif: If we make an enum that just duplicates what's already stored in the permissions table. How about storing a column for permission name and permission description, then making a function that filters on permission name?

davistardif:
Should be ready for a review

donut/modules/directory_search/helpers.py
```diff
@@ -12,7 +12,7 @@ def get_hidden_fields(viewer_name, viewee_id):
     """
     if viewer_name is not None:
         is_me = get_user_id(viewer_name) == viewee_id
-        if is_me or Permissions.ADMIN in get_permissions(viewer_name):
+        if is_me or 3 in get_permissions(viewer_name):
```
calebsander: Why use a magic number here?

donut/modules/core/routes.py
```diff
@@ -2,7 +2,7 @@
 import json
 from flask import jsonify, redirect
 
-from donut.auth_utils import get_user_id
+from donut import auth_utils
```
calebsander, line 5: Why change this file?

donut/auth_utils.py
```diff
@@ -318,67 +319,22 @@ def login_redirect():
...
+    if not user_id: return []
+    positions = groups.get_positions_held(user_id)
+    query = """
+    SELECT permission_id FROM position_permissions WHERE pos_id in 
+    (%s)""" % (', '.join(['%s'] * len(positions)))
     with flask.g.pymysql_db.cursor() as cursor:
-        cursor.execute(query, username)
+        cursor.execute(query, (positions))
         result = cursor.fetchall()
     return list(row['permission_id'] for row in result)
```
calebsander: Can this be made into a `set` since it's never serialized into `flask.session`?

donut/modules/auth/routes.py
```diff
@@ -20,10 +20,6 @@ def login_submit():
         if user_id is not None:
             permissions = auth_utils.get_permissions(username)
```
calebsander: This is now unused

donut/auth_utils.py
```diff
@@ -318,67 +319,22 @@ def login_redirect():
...
     """
+    user_id = get_user_id(username)
+    if not user_id: return []
+    positions = groups.get_positions_held(user_id)
+    query = """
+    SELECT permission_id FROM position_permissions WHERE pos_id in 
+    (%s)""" % (', '.join(['%s'] * len(positions)))
     with flask.g.pymysql_db.cursor() as cursor:
-        cursor.execute(query, username)
+        cursor.execute(query, (positions))
```
calebsander: Parens unnecessary

donut/auth_utils.py
calebsander: Also, `list(...)` can be shortened to `[...]`

calebsander (dismissed):

donut/modules/core/routes.py
davistardif: There was a circular import issue -- through a chain of imports, auth_utils imports this file. Then this file looked for a function in auth_utils which hasn't actually been defined yet. If we just import the whole module it works fine and only looks for the function at runtime.

donut/auth_utils.py
davistardif: Good point, will change

donut/modules/auth/routes.py
davistardif: Good catch, will remove

donut/modules/directory_search/helpers.py
davistardif: I don't really have a great solution for that. I don't think we should store a massive enum of all permissions, because then we are just replicating the database with python. Does it make sense for each module that uses permissions to maintain its own enum? Either way I was just stuffing that in there as a placeholder to make sure the directory search tests still pass

calebsander: Do we even need a permissions table? It seems like we could just store all the integer permissions values in a `IntEnum` and insert those values directly into `permission_id` columns.

davistardif: I suppose that's possible too. Let's discuss with the whole team on Friday.

davistardif:
I created a wiki page with documentation:
https://github.com/ASCIT/donut/wiki/Permissions

donut/modules/directory_search/helpers.py
davistardif: We decided at the meeting to have a global enum for default (i.e. non module specific) permissions and having modules define permissions enums as needed.

calebsander: Hmm, so do we still need the permissions tables?

majesticfish (approved):
LGTM

PR #80, opened 2018-10-09
Bodfeedback

rlin0:
### Summary
- Bodfeedback form submission, summary page for feedback, and email notifications for people subscribed


rlin0:
@davistardif

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    status = 'new_msg'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (data['subject'], status))
+    # grab the complaint id that we just inserted
+    query = 'SELECT LAST_INSERT_ID()'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query)
+        res = cursor.fetchone()
+        complaint_id = res['LAST_INSERT_ID()']
```
calebsander: Can just use `cursor.lastrowid` after you execute the insertion

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+
+def get_link(complaint_id):
+    """
+    Gets a (fully qualified) link to the view page for this complaint id
+    """
+    query = 'SELECT uuid FROM bod_complaint_info WHERE complaint_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchone()
+        if res and 'uuid' in res:
```
calebsander: Not sure you need to check whether `'uuid' in res`; if a row matches, it should definitely have a `uuid` value

sql/bodfeedback.sql
```diff
@@ -0,0 +1,33 @@
+DROP TABLE IF EXISTS `bod_complaint_info`;
```
calebsander: Don't think backticks are necessary

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+            uuid = res['uuid']
+        else:
+            uuid = None
```
calebsander: Could just `return None` here

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    Returns a dict with the following fields: course, status
+    """
+    fields = ['subject', 'status']
+    query = 'SELECT ' + ', '.join(
+        fields) + ' FROM bod_complaint_info WHERE complaint_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchone()
+    return res if res else None
```
calebsander: If `res` is false, shouldn't is already be `None`?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+    or an empty list if complaint_id is invalid
+    """
+    query = 'SELECT email FROM bod_complaint_emails WHERE complaint_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchall()
+    emails = []
+    for row in res:
+        emails.append(row['email'])
+    return emails
```
calebsander: `return [row['email'] for row in res]`

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,278 @@
...
+def get_all_fields(complaint_id):
+    '''
+    Returns a dict with emails, messages, subject, status
+    Returns None if complaint_id is invalid
+    '''
+    data = {}
+    data['emails'] = get_emails(complaint_id)
+    data['messages'] = get_messages(complaint_id)
+    data['subject'] = get_subject(complaint_id)
+    data['status'] = get_status(complaint_id)
```
calebsander: Would be cleaner to use a dict literal to initialize `data`

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,140 @@
...
+# submit feedback form
+@blueprint.route('/bodfeedback/submit', methods=['POST'])
+def bodfeedback_submit():
+    fields = ['name', 'email', 'subject', 'msg']
+    required = ['subject', 'msg']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    for field in required:
+        if (data[field] == ""):
```
calebsander: Parens unnecessary

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,140 @@
...
+    if data['email'] != "":
+        helpers.send_confirmation_email(data['email'], complaint_id)
+    flask.flash('Success. Link: ' + helpers.get_link(complaint_id))
+    return flask.redirect(flask.url_for('bodfeedback.bodfeedback'))
+
+
+# api endpoint with all visible data
+@blueprint.route('/1/bodfeedback/view/<uuid:id>')
+def bodfeedback_api_view_complaint(id):
+    if not (helpers.get_id(id)):
```
calebsander: Parens unnecessary

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,140 @@
...
+        return flask.render_template("404.html")
+    complaint_id = helpers.get_id(id)
+    #pack all the data we need into a dict
+    return flask.jsonify(helpers.get_all_fields(complaint_id))
+
+
+# view a complaint
+@blueprint.route('/bodfeedback/view/<uuid:id>')
+def bodfeedback_view_complaint(id):
+    if not (helpers.get_id(id)):
```
calebsander: Parens unnecessary

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+# remove an email from this complaint
+@blueprint.route('/1/bodfeedback/removeEmail/<uuid:id>', methods=['POST'])
+def bodfeedback_remove_email(id):
+    complaint_id = helpers.get_id(id)
+    if not complaint_id:
+        flask.abort(400)
+        return
+    data = []
```
calebsander: This line has no effect

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+		<button class="btn btn-primary" id="markUnread">Mark as unread</button>
+	    </div>
+	</div>
+	<div class="col-md-4">
+	    <div class="panel panel-default">
+		<div class="panel-heading">
+		    <h4>Add an email</h4>
+		</div>
+		<div class="panel-body">
+			<form method="POST" id="add-email" action="/">
```
calebsander: `action` and `method` are not used; should be omitted

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+{{ super() }}
+<script>
+ 
+ $(document).ready(function() {
+     {% if complaint['status'] == 'new_msg' %}
+     $('#markUnread').hide();
+     {% else %}
+     $('#markRead').hide();
+     {% endif %} 
+     $('#add-email').on('submit', function(e) {
```
calebsander: Can use `.submit(function(e) {})` instead

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+	     url: "{{ url_for('bodfeedback.bodfeedback_add_msg', id=complaint['uuid']) }}",
+	     data: $('#addMessage').serialize(),
+	     type: 'POST', 
+	     success: function(){
+		 $('#addMessage')[0].reset();
+	         location.reload();
+	     }
+	 }); 
+     });
+     $('#markRead').on('click', function() {
```
calebsander: Likewise, `.click(function() {})`

donut/modules/bodfeedback/templates/summary.html
```diff
@@ -0,0 +1,26 @@
...
+	{% endfor %}
+    </table>
+</div>
+
+{% endblock %}
+
+{% block scripts %}
+{{super()}}
+
+{% endblock %}
```
calebsander, line 20: This `block` can be removed

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,129 @@
...
+from donut.modules.bodfeedback import helpers
+from donut.testing.fixtures import client
+from datetime import datetime
+
+
+def test_get_link(client):
+    assert helpers.get_link(
+        1) == flask.url_for('home') + 'bodfeedback/view/sample_uuid'
+    assert helpers.get_link(500) is None
+    return
```
calebsander: Trailing `return` statements are unnecessary

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def test_get_messages(client):
+    messages = helpers.get_messages(1)
+    assert messages[0]['time'] == datetime(2018, 1, 1, 0, 0)
+    assert messages[1]['time'] == datetime(2018, 1, 2, 0, 0)
+    assert messages[0]['message'] == 'Sample Message'
+    assert messages[1]['message'] == 'Sample Message 2'
+    assert messages[0]['poster'] == 'Davis'
+    assert messages[1]['poster'] == 'Davis'
```
calebsander: Would be clearer to do `assert messages == [{...}, {...}]`, and similarly in other assertions

calebsander (changes requested):

rlin0:
@calebsander changes made

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,259 @@
...
+    returns False if complaint_id is invalid
+    '''
+    query = """
+    UPDATE bod_complaint_info SET status = 'new_msg' WHERE complaint_id = %s
+    """
+    if get_status(complaint_id) is None:
+        return False
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        cursor.execute("COMMIT;")  #not sure why this doesn't autocommit
```
koivyr: Nitpick: Put the side comment above the line of code.

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,137 @@
...
+        flask.abort(400)
+        return
+    return 'Success'
+
+
+# mark a complaint unread
+@blueprint.route('/1/bodfeedback/markUnread/<uuid:id>')
+def bodfeedback_mark_unread(id):
+    #authenticate
+
```
koivyr: Nit: comment followed by empty line.

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,137 @@
...
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    if data['email'] == "":
+        flask.abort(400)
+        return
+    helpers.add_email(complaint_id, data['email'])
+    return flask.jsonify({'email': data['email']})
+
+
+# remove an email from this complaint
```
koivyr: More nit: in general, comments start with a capital letter per the coding guidelines.

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,259 @@
...
+def add_msg(complaint_id, message, poster, notification=True):
+    '''
+    Adds a message to a complaint in the database
+    and updates status of complaint to 'new_msg'
+    if poster is None or an empty string, it will be replaced with
+    "(anonymous)"
+    If complaint_id is invalid, returns False
+    '''
+    if not get_subject(complaint_id): return False
+    #add the message
```
koivyr: Nit: Either add a space after the # or not, but be consistent

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,137 @@
...
+        return
+    fields = ['message', 'poster']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    if data['message'] == "":
+        flask.abort(400)
+        return
+    helpers.add_msg(complaint_id, data['message'], data['poster'])
+    return flask.jsonify({
```
calebsander: Why not just `return flask.jsonify(data)`?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,259 @@
...
+    return res
+
+
+def get_summary(complaint_id):
+    """
+    Returns a dict with the following fields: course, status
+    """
+    fields = ['subject', 'status']
+    query = 'SELECT ' + ', '.join(
+        fields) + ' FROM bod_complaint_info WHERE complaint_id = %s'
```
calebsander: Why not just write `SELECT subject, status FROM ...` if `fields` is constant?

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,137 @@
...
+        return
+    fields = ['email']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    if data['email'] == "":
+        flask.abort(400)
+        return
+    helpers.add_email(complaint_id, data['email'])
+    return flask.jsonify({'email': data['email']})
```
calebsander: Can't this also just be `return flask.jsonify(data)`?

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,137 @@
...
+    if not complaint_id:
+        flask.abort(400)
+        return
+    data = flask.request.form.getlist('emails')
+    if data == []:
+        flask.abort(400)
+        return
+    for em in data:
+        helpers.remove_email(complaint_id, em)
+    return flask.jsonify({'email': data[0]})
```
calebsander: Why just return the first email? Does any data really need to be sent to the client?

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+    <div class="row">
+	<div class="col-md-8">
+	    <div class="panel panel-default">
+		<div class="panel-heading">
+		    <h3>Feedback Summary</h3>
+		</div>
+		<div class="panel-body" id="summary-panel">
+		    <p>Subject: {{complaint['subject']}} </p>
+		    <p>Status: {{complaint['status']}} </p>
+		    {% if complaint['emails']|length == 0 %}
```
calebsander: Could instead check `if complaint['emails']`

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+	 e.preventDefault();
+	 $.ajax({
+	     url: "{{ url_for('bodfeedback.bodfeedback_remove_email', id=complaint['uuid']) }}",
+             data: $('#remove-email').serialize(),
+	     type: 'POST', 
+	     success: function(){
+	         location.reload();
+	     }     	 
+	 }); 
+     });
```
calebsander: Could just use an HTML form--there isn't any need for a custom JavaScript event handler

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,167 @@
...
+     $('#markRead').click(function() {
+	 $.ajax({
+	     url: "{{ url_for('bodfeedback.bodfeedback_mark_read', id=complaint['uuid']) }}",
+	     type: 'GET',
+	     success: function(){
+		 $('#markRead').hide();
+		 $('#markUnread').show();
+	     },
+	     error: function(response){
```
calebsander: The first argument passed to the error handler is an `XMLHttpRequest` object; it doesn't really make sense to append it to the page

sql/bodfeedback.sql
```diff
@@ -0,0 +1,33 @@
...
+CREATE TABLE bod_complaint_info (
+  complaint_id int(11) NOT NULL AUTO_INCREMENT,
+  subject varchar(50) DEFAULT NULL,
+  status varchar(36) DEFAULT NULL,
+  uuid varchar(36) NOT NULL,
+  PRIMARY KEY (complaint_id),
+  UNIQUE KEY (uuid)
+);
+
+DROP TABLE IF EXISTS bod_complaint_messages;
```
calebsander: I don't think the order of these `DROP TABLE` statements will work, since these second and third tables have `FOREIGN KEY` constraints that depend on the first table. I think you have to drop `bod_complaint_messages` and `bod_complaint_emails` first.

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,139 @@
...
+    assert helpers.get_emails(2) == []
+    assert helpers.get_emails(500) == []
+
+
+def test_get_all_fields(client):
+    fields = helpers.get_all_fields(1)
+    assert fields['emails'] == ['test@example.com', 'test2@example.com']
+    assert fields['messages'][0]['message'] == 'Sample Message'
+    assert fields['subject'] == 'Sub1'
+    assert fields['status'] == 'new_msg'
```
calebsander: I would use a single `assert fields == {...}` here

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,139 @@
...
+    assert helpers.get_all_fields(500) is None
+
+
+def test_get_new_posts(client):
+    posts = helpers.get_new_posts()
+    assert len(posts) == 1
```
calebsander: This length assertion is redundant

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,139 @@
...
+    res = helpers.get_all_fields(complaint_id)
+    expected = {
+        'subject': 'Sub1',
+        'status': 'new_msg',
+        'emails': ['joe@example.com', 'joey@example.com']
+    }
+    for key in expected:
+        assert expected[key] == res[key]
+    assert res['messages'][0]['message'] == 'Sample Message 3'
+    assert res['messages'][0]['poster'] == 'Joe Schmo'
```
calebsander: Would be simpler with a single `assert res == {...}`. Also, it is convention to use `assert actual == expected`; if you switch the order, the error messages will be confusing.

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,139 @@
...
+    expected_messages = [{
+        'message': 'This course is fun',
+        'poster': 'Davis',
+    }, {
+        'message': 'Test message',
+        'poster': 'Test user',
+    }]
+    for i in range(len(expected_messages)):
+        for key in expected_messages[i]:
+            assert expected_messages[i][key] == messages[i][key]
```
calebsander: Again, could just use a single assertion: `assert messages == expected_messages`. The way this is currently written, it will pass if `messages` has additional messages or extra keys not present on `expected_messages`. Not sure if that is intentional.

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,139 @@
...
+    helpers.add_msg(2, 'Anonymous message', '', False)
+    messages = helpers.get_messages(2)
+    assert len(messages) == 3
+    assert messages[2]['poster'] == '(anonymous)'
```
calebsander: Should probably check the whole value of `messages`. Could do something like `assert messages == expected_messages + [{'message': 'Anonymous message', 'poster': '(anonymous)'}]`

calebsander (changes requested):

tests/modules/bodfeedback/test_bodfeedback.py
rlin0: It was intentional, because add_msg also includes datetime so get_messages has date_time but I don't know the expected datetime. If I put 'date': '.' would it interpret the . as any character (as in regular expressions)? I could try that

calebsander: You could do something like `messages_without_date = [{k: v for (k, v) in message.items() if k != 'date'} for message in messages]` and then `assert expected_messages == messages_without_date`

donut/modules/bodfeedback/routes.py
rlin0: Not sure why it was just the 1st email. Yes or the page does not get reloaded with the updated emails

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,132 @@
...
+        return
+    fields = ['message', 'poster']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    if data['message'] == "":
+        flask.abort(400)
+        return
+    helpers.add_msg(complaint_id, data['message'], data['poster'])
+    return flask.redirect('/bodfeedback/view/' + str(id))
```
calebsander: Can this be done with a `flask.url_for(..)`?

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,132 @@
...
+    if not complaint_id:
+        flask.abort(400)
+        return
+    data = {}
+    data['email'] = flask.request.form.get('email')
+    if data['email'] == "":
+        flask.abort(400)
+        return
+    helpers.add_email(complaint_id, data['email'])
+    return flask.redirect('/bodfeedback/view/' + str(id))
```
calebsander: Can this be done with a `flask.url_for(..)`?

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,132 @@
...
+    if not complaint_id:
+        flask.abort(400)
+        return
+    data = flask.request.form.getlist('emails')
+    if data == []:
+        flask.abort(400)
+        return
+    for em in data:
+        helpers.remove_email(complaint_id, em)
+    return flask.redirect('/bodfeedback/view/' + str(id))
```
calebsander: Can this be done with a `flask.url_for(..)`?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+    return cursor.fetchone()['complaint_id']
+
+
+def get_messages(complaint_id):
+    '''
+    Returns timestamps, posters, messages, and message_id's on this complaint
+    in ascending order of timestamp
+    '''
+    query = """
+    SELECT time, poster, message, message_id FROM bod_complaint_messages WHERE complaint_id = %s ORDER BY time
```
joshc: what's the point of a multiline string if you don't split the string into multiple lines 🗡

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+
+def get_summary(complaint_id):
+    '''
+    Returns a dict with the following fields: subject, status
+    '''
+    query = 'SELECT subject, status FROM bod_complaint_info WHERE complaint_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchone()
+    return res
```
joshc: should initialize `res` to None otherwise you may potentially never set `res`

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+    '''
+    query = """
+    UPDATE bod_complaint_info SET status = 'new_msg' WHERE complaint_id = %s
+    """
+    if get_status(complaint_id) is None:
+        return False
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        # Not sure why this doesn't autocommit
+        cursor.execute("COMMIT;")
```
joshc: this could be why [https://stackoverflow.com/a/35621700](link). @calebsander do you think we should enable autocommit? or otherwise this is probably fine.

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+def get_emails(complaint_id):
+    '''
+    Returns a list of subscribed emails for this complaint (which may be empty)
+    or an empty list if complaint_id is invalid
+    '''
+    query = 'SELECT email FROM bod_complaint_emails WHERE complaint_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchall()
+    return [row['email'] for row in res]
```
joshc, line 206: same potential issue if `res` is not set

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def get_new_posts():
+    '''
+    Returns all posts with status 'new_msg' and their associated list
+    of messages. Will be an array of dicts with keys complaint_id, subject,
+    status, uuid, message, poster, time
+    Note that message and poster refer to the latest comment on this complaint
+    '''
+    query = """SELECT post.complaint_id AS complaint_id, post.subject AS subject, post.status AS status,
```
joshc: lengthy line

donut/modules/bodfeedback/permissions.py
```diff
@@ -0,0 +1,9 @@
+import enum
```
joshc, line 1: @davistardif this is the first feature since permissions was birthed to use permissions. can you check that it is fine under your design :)

donut/modules/bodfeedback/templates/bodfeedback.html
```diff
@@ -0,0 +1,65 @@
+{% extends "layout.html" %}
+{% block page %}
+
+<div class="container theme-showcase" role="main">
+
+    <!-- Main jumbotron for a primary marketing message or call to action -->
+    <div class="jumbotron">
+        <img src="{{  url_for('static', filename='images/newdonut.png')  }}" alt="d(O)nut" title="d(O)nut" style="margin: auto; display: block">
```
joshc: see our current home page for @Thought-Weaver's updated logo image. as a side note, any design comments @Thought-Weaver

donut/modules/bodfeedback/templates/complaint.html
joshc: has this been resolved?

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,139 @@
...
+	</div>
+    </div>
+</div>
+
+{% endblock %}
+{% block scripts %}
+{{ super() }}
+<script>
+ 
+ $(document).ready(function() {
```
joshc, line 101: since this a somewhat significant amount of code might be cleaner to have it not inline. also spacing seems a little off.

donut/templates/includes/header.html
```diff
@@ -27,6 +27,7 @@
                         <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">User Services <span class="caret"></span></a>
                         <ul class="dropdown-menu" role="menu">
                             <li><a href="#">ARC Feedback</a></li>
+			    <li><a href="{{ url_for('bodfeedback.bodfeedback') }}">BoD Feedback</a></li>
```
joshc: previously the bod feedback form was linked under ASCIT -> Suggestions. If you think it makes more sense to put it here, you should remove that one. otherwise remove this and link that

donut/modules/bodfeedback/templates/bodfeedback.html
```diff
@@ -0,0 +1,65 @@
...
+	<p>Ever wanted to suggest something to the Board of Director's, but didn't want to say it directly / didn't know when BoD meetings were? Ever wanted to comment about how an event was run, or are curious about how to get involved?
+
+Well then, simply fill out the form below. The ASCIT BoD would love to hear all of your suggestions</p>
+
+	<p>Fill in your name and email address if you are willing to have the BoD contact
+  you with follow up questions; this will obviously make your comment
+  non-anonymous, but we do promise to keep comments strictly
+  confidential. </p>
+
+	<p>To: Board of Directors</p>
```
joshc: this looks kind of weird lol. maybe remove it

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,255 @@
...
+    '''
+    Sends an email to [email] which should be an
+    email address of a user as a string
+    '''
+    msg = email_templates.added_message.format(get_link(complaint_id))
+    subject = "Received BoD Feedback"
+    email_utils.send_email(email, msg, subject)
+
+
+def register_complaint(data, notification=True):
```
joshc, line 17: do you send an email to bod to tell them a complaint has been registered? what are reasons for or against this? what is the behavior of bodfeedback form on the legacy website? is the desired behavior for bod for them to check the listings during their weekly meetings? 🥇

joshc (changes requested):
Didn't comment too much on the code since I assume Ziyan/Cleb caught most to all of the errors. There are some changes that should be considered generally, and also some specific comments. Looping in @Thought-Weaver and @davistardif to take a quick look at design and permissions respectively before approving the merge 💯

donut/modules/bodfeedback/helpers.py
calebsander: Or just `return cursor.fetchone()`

donut/modules/bodfeedback/helpers.py
calebsander: We already have `autocommit` turned on: https://github.com/ASCIT/donut/blob/865dccd4c190078a7d856cb13227007d014808aa/donut/__init__.py#L85

donut/modules/bodfeedback/helpers.py
calebsander: Unless an exception occurs in the `with` block, `res` should always be set

donut/modules/bodfeedback/templates/complaint.html
calebsander: No

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,168 @@
...
+        data[field] = flask.request.form.get(field)
+    for field in required:
+        if not data[field]:
+            flask.flash('Please fill in all required fields (marked with *)',
+                        'error')
+            return flask.redirect(flask.url_for('bodfeedback.bodfeedback'))
+    complaint_id = helpers.register_complaint(data)
+    flask.flash(
+        Markup('Success: <a href="' + helpers.get_link(complaint_id) +
+               '">View Complaint</a>'))
```
koivyr: Maybe add an extra note telling them to save this url if they want to keep track of this complaint?

donut/modules/bodfeedback/email_templates.py
```diff
@@ -0,0 +1,7 @@
+added_message = \
+"""
+We have received BoD feedback that you're subscribed to.
+You can check the status of your feedback or unsubscribe here: {0}\n
```
calebsander: Use actual newline character instead of `\n`

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    if not (data and data['subject'] and data['msg']): return False
+    # Register complaint
+    query = """
+    INSERT INTO bod_complaint_info (subject, status, uuid)
+    VALUES (%s, %s, UUID())
+    """
+    status = 'new_msg'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (data['subject'], status))
+    complaint_id = cursor.lastrowid
```
calebsander: This line should go inside the `with` block since you are still using `cursor`

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    # Add email to db if applicable
+    if data['email']:
+        emails = [x.strip() for x in data['email'].split(',')]
+        for e in emails:
+            add_email(complaint_id, e, False)
```
calebsander: This iterates over the list of emails twice. Would be better to do something like:
```python
for email in data['emails'].split(','):
    add_email(complaint_id, email.strip(), False)
```

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    if not get_subject(complaint_id): return False
+    query = """
+    INSERT INTO bod_complaint_emails (complaint_id, email)
+    VALUES (%s, %s)
+    """
+    try:
+        with flask.g.pymysql_db.cursor() as cursor:
+            cursor.execute(query, (complaint_id, email))
+        if notification:
+            send_update_email(email, complaint_id)
```
calebsander: Does this `send_update_email()` need to be guarded by the `pymysql.err.IntegrityError` check?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    '''
+    Returns the complaint_id associated with a uuid
+    or false if the uuid is not found
+    '''
+    query = 'SELECT complaint_id FROM bod_complaint_info WHERE uuid = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, str(uuid))
+        if not cursor.rowcount:
+            return False
+    return cursor.fetchone()['complaint_id']
```
calebsander: `cursor.fetchone()` needs to go inside the `with` block

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    in ascending order of timestamp
+    '''
+    query = """
+    SELECT time, poster, message, message_id FROM bod_complaint_messages 
+    WHERE complaint_id = %s ORDER BY time
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchall()
+    if not (res and 'message_id' in res[0]):
```
calebsander: Shouldn't `'message_id' in res[0]` always be true?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    res = get_summary(complaint_id)
+    return res['status'] if res else None
+
+
+def mark_read(complaint_id):
+    '''
+    Sets the status of this complaint to 'read'
+    returns False if complaint_id is invalid
+    '''
+    query = "UPDATE bod_complaint_info SET status = 'read' WHERE complaint_id = %s"
```
calebsander: `query` doesn't need to be instantiated if the function returns early

calebsander: (And in `mark_unread()`)

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, complaint_id)
+
+
+def mark_unread(complaint_id):
+    '''
+    Sets the status of this complaint to 'new_msg'
+    returns False if complaint_id is invalid
+    '''
+    query = "UPDATE bod_complaint_info SET status = 'new_msg' WHERE complaint_id = %s"
```
calebsander: Should these statuses (stati?), e.g. `new_msg` and `read` be defined as enums somewhere?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns a dict with emails, messages, subject, status
+    Returns None if complaint_id is invalid
+    '''
+    data = {
+        'emails': get_emails(complaint_id),
+        'messages': get_messages(complaint_id),
+        'subject': get_subject(complaint_id),
+        'status': get_status(complaint_id)
+    }
+    return data if data['subject'] else None
```
calebsander: This check should happen before all the functions are called on `complaint_id`

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+    post.status AS status, post.uuid AS uuid, comment.message AS message, 
+    comment.poster AS poster, comment.time AS time 
+    FROM bod_complaint_info post
+    INNER JOIN bod_complaint_messages comment
+    ON comment.complaint_id = post.complaint_id
+    INNER JOIN (
+    SELECT complaint_id, max(time) AS time 
+    FROM bod_complaint_messages GROUP BY complaint_id
+    ) maxtime
+    ON maxtime.time = comment.time AND maxtime.complaint_id = comment.complaint_id
```
calebsander, line 241: Can't this be done with `NATURAL JOIN` instead? Would be a lot more readable

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Submit feedback form
+@blueprint.route('/bodfeedback/submit', methods=['POST'])
+def bodfeedback_submit():
+    fields = ['name', 'email', 'subject', 'msg']
+    required = ['subject', 'msg']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
```
calebsander, line 27: Why copy this into `data`? Can't the `form` be used as a `dict` instead?

donut/modules/bodfeedback/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+        cursor.execute(query, complaint_id)
+        res = cursor.fetchone()
+        if not res:
+            return None
+        uuid = res['uuid']
+    return flask.url_for(
+        'bodfeedback.bodfeedback_view_complaint', id=uuid, _external=True)
+
+
+def get_id(uuid):
```
calebsander, line 120: Why keep a separate `uuid` and `id` for each complaint? Can't the `uuid` just be used as the primary key instead?

sql/bodfeedback.sql
```diff
@@ -0,0 +1,32 @@
+DROP TABLE IF EXISTS bod_complaint_emails;
+DROP TABLE IF EXISTS bod_complaint_messages;
+DROP TABLE IF EXISTS bod_complaint_info;
+
+CREATE TABLE bod_complaint_info (
+  complaint_id int(11) NOT NULL AUTO_INCREMENT,
+  subject varchar(50) DEFAULT NULL,
+  status varchar(36) DEFAULT NULL,
+  uuid varchar(36) NOT NULL,
```
calebsander: Use `BINARY(16)` with `UUID_TO_BIN()` and `BIN_TO_UUID()` as described here: www.mysqltutorial.org/mysql-uuid/

donut/modules/bodfeedback/templates/complaint.html
```diff
@@ -0,0 +1,139 @@
...
+				<textarea class="form-control" name="message" rows="10" required></textarea>
+			    </div>
+			    <input class="btn btn-primary" id="submitBtn" type="submit" value="Submit" />
+			</form>
+		    </div>
+		</div>
+	    </div>
+	</div>
+    </div>
+</div>
```
calebsander, line 94: Your HTML files are mixing tabs and spaces. Please pick one or the other

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,168 @@
...
+        'complaint.html', complaint=complaint, perms=namedperms)
+
+
+# Add a message to this post
+@blueprint.route('/1/bodfeedback/add/<uuid:id>', methods=['POST'])
+def bodfeedback_add_msg(id):
+    complaint_id = helpers.get_id(id)
+    if not complaint_id:
+        flask.abort(400)
+        return
```
calebsander, line 75: `return` is unreachable here (and below)

donut/modules/bodfeedback/routes.py
```diff
@@ -0,0 +1,168 @@
...
+            flask.session['username'], bod_permissions.ADD_REMOVE_EMAIL):
+        flask.abort(403)
+        return
+    complaint_id = helpers.get_id(id)
+    if not complaint_id:
+        flask.abort(400)
+        return
+    data = flask.request.form.getlist('emails')
+    if data:
+        for em in data:
```
calebsander: If `data` is empty, the `for` loop should never run anyways, so I think `if data` is redundant

tests/modules/bodfeedback/test_bodfeedback.py
```diff
@@ -0,0 +1,172 @@
+import flask
+from donut.modules.bodfeedback import helpers
+from donut.testing.fixtures import client
+from datetime import datetime
+
+
+def test_get_link(client):
+    assert helpers.get_link(
+        1) == flask.url_for('home') + 'bodfeedback/view/sample_uuid'
```
calebsander: Can't this be done with a `flask.url_for()` call rather than building the path yourself?

sql/bodfeedback.sql
rlin0: UUID_TO_BIN() is only in MySQL not in MariaDB I think.
https://mariadb.com/kb/en/library/function-differences-between-mariadb-103-and-mysql-80/

calebsander: Use `UNHEX(REPLACE(UUID(), '-', ''))` to generate a UUID and `HEX(uuid)` to convert it back to hexadecimal

joshc (dismissed):

donut/modules/bodfeedback/routes.py
rlin0: I can't find an analogous cursor.lastrowid to get the most recently generated UUID (in this line https://github.com/ASCIT/donut/blob/bodfeedback/donut/modules/bodfeedback/helpers.py#L32

donut/modules/bodfeedback/helpers.py
rlin0: I can't find an analogous cursor.lastrowid to get the most recently generated UUID (in this line https://github.com/ASCIT/donut/blob/bodfeedback/donut/modules/bodfeedback/helpers.py#L32

PR #82, opened 2018-10-13
Added Tab to add position holders

majesticfish:
### Summary
- Added a new tab to the positions administrate tab to add people to position holders
- Imported some of @calebsander code for a search box that looks up users
- When looking up user, they will come up as ("Name, Grad_year")

### Test Plan
- The e2e test locally seems to work


donut/modules/groups/routes.py
```diff
@@ -37,6 +37,7 @@ def get_positions():
             and validate_int(form["group_id"]),
             validate_exists(form, "pos_name"),
         ]
+        print(str(form))
```
calebsander: Remove `print` statement

donut/modules/groups/routes.py
```diff
@@ -78,15 +79,14 @@ def get_pos_holders(pos_id):
 @blueprint.route("/1/positions/<int:pos_id>/", methods=["POST"])
 def create_pos_holder(pos_id):
     form = flask.request.form
+    print(form)
```
calebsander: Remove `print` statement

donut/static/js/groups/positions.js
```diff
@@ -179,16 +204,75 @@ $(document).ready(function() {
...
+                    setUpForms();
+                    alert("Assigned Position!") 
+                }
+            }
+        });
+    });
+
+    // script for updating the seach bar for looking up people
+    // taken from csanders directory search code
+    var nameInput = $('input#name'), nameSearch = $('ul#name-search')
```
calebsander: If possible, it would be great to just include `search-directory.js` so we don't have to maintain this code in 2 places

calebsander (changes requested):

donut/modules/groups/routes.py
```diff
@@ -78,15 +79,14 @@ def get_pos_holders(pos_id):
...
     validations = [
-        validate_exists(form, "user_id") and validate_int(form["user_id"]),
-        validate_exists(form, "start_date")
-        and validate_date(form["start_date"]),
-        validate_exists(form, "end_date") and validate_date(form["end_date"])
+        validate_exists(form, "userId"),
+        validate_exists(form, "start_date"),
+        validate_exists(form, "end_date")
```
joshc: why are the checks that start_date or end_date is a valid date removed?

donut/modules/groups/routes.py
```diff
@@ -78,15 +79,14 @@ def get_pos_holders(pos_id):
 @blueprint.route("/1/positions/<int:pos_id>/", methods=["POST"])
 def create_pos_holder(pos_id):
     form = flask.request.form
+    print(form)
     validations = [
-        validate_exists(form, "user_id") and validate_int(form["user_id"]),
-        validate_exists(form, "start_date")
-        and validate_date(form["start_date"]),
-        validate_exists(form, "end_date") and validate_date(form["end_date"])
+        validate_exists(form, "userId"),
```
joshc: why did user_id -> userId? try to stay consistent (underscores in python)

donut/templates/campus_positions.html
```diff
@@ -84,6 +87,44 @@ <h1 id="pagetitle" > The Ascit Position Service </h1>
                         <button id="delPosBtn" class="btn btn-default">Submit</button>
                     </form>     
                 </div>
+                <div id="assignHold" class="tab-pane fade in">
+                    <form id="holdForm">
+                        <div class="form-group">
+                            <label class="control-label col-sm-2" for="groupPosHold">
```
joshc, line 94: why are group/position and name/start/end not aligned?

donut/templates/campus_positions.html
```diff
@@ -84,6 +87,44 @@ <h1 id="pagetitle" > The Ascit Position Service </h1>
...
+                            <label class="control-label col-sm-2" for="posIdHold">
+                                Position: </label>
+                            <select class="form-control" id="posIdHold" 
+                                                            name="pos_id">
+                                <option> Pick a position </option>
+                            </select>
+                        </div>
+                        <div class='form-group'>
+                            <label for='name'>Name: </label>
+                            <input class='form-control' id='name' name='name'
```
joshc, line 112: add autocomplete="off" to disable browser autocomplete

donut/static/js/groups/positions.js
majesticfish: Done

majesticfish:
I think i've addressed all comments, see lastest commit

donut/static/js/groups/positions.js
```diff
@@ -130,27 +132,50 @@ function changePosition(posName) {
...
             }
         }
-    })
+    });
 }
 
-// Setting up triggers for administration tasks
+var MIN_SEARCH_LENGTH = 3
+var SEARCH_TIMEOUT = 100 //ms
+var searchToken
```
calebsander: Unneeded variables, I think

donut/static/js/groups/positions.js
```diff
@@ -179,16 +204,35 @@ $(document).ready(function() {
             }
         });
     });
-}); 
+    $('#submitPosHoldBtn').click(function(e) {
+        e.preventDefault();
+        $.ajax({
+            type: 'POST',
+            url: '/1/positions/' + $("#posIdHold").val() + '/',
```
calebsander, line 207: Don't think trailing slash is needed

donut/templates/campus_positions.html
```diff
@@ -93,3 +135,23 @@ <h1 id="pagetitle" > The Ascit Position Service </h1>
...
+{% block scripts %}
+    {{ super()  }}
+    <script src='{{ url_for("static", filename="js/search-directory.js") }}'></script>
+    <script>
+        $(document).ready(function() {
+            var nameInput = $('input#name'), nameSearch = $('ul#name-search')
+            attachDirectorySearch(nameInput, nameSearch, function(user){
+                return $('<li>').addClass('list-group-item').append($('<a>')
+                        .text(user.full_name + ", " + user.graduation_year)
+                        .click({name: user.full_name, user_id: user.user_id},
```
calebsander: I don't think you need to bind the `user` fields here; just use `user` in the callback instead of `event.data`

donut/static/js/groups/positions.js
majesticfish: Its not _needed_ but flask does complain if its not there:
`
"'A request was sent to this URL (http://127.0.0.1:9001/1/positions/2) but a redirect was issued automatically by the routing system to "http://127.0.0.1:9001/1/positions/2/".  The URL was defined with a trailing slash so Flask will automatically redirect to the URL with the trailing slash if it was accessed without one.  Make sure to directly send your POST-request to this URL since we can\'t make browsers or HTTP clients redirect with form data reliably or without user interaction.\n\nNote: this exception is only raised in debug mode'"
`

donut/modules/groups/routes.py
```diff
@@ -79,14 +79,13 @@ def get_pos_holders(pos_id):
...
         and validate_date(form["start_date"]),
-        validate_exists(form, "end_date") and validate_date(form["end_date"])
+        validate_exists(form, "end_date") and validate_date(form["end_date"]),
     ]
     if not all(validations):
         return jsonify({'success': False})
-    helpers.create_position_holder(
-        int(pos_id), int(form["user_id"]), form["start_date"],
-        form["end_date"])
+    helpers.create_position_holder(pos_id, form["user_id"], form["start_date"],
```
calebsander: `form["user_id"]` should really be cast to an `int` so the SQL statement is not assigning a string to an integer. I would add `validate_int` back in too.

donut/modules/groups/routes.py
```diff
@@ -79,14 +79,13 @@ def get_pos_holders(pos_id):
 def create_pos_holder(pos_id):
     form = flask.request.form
     validations = [
-        validate_exists(form, "user_id") and validate_int(form["user_id"]),
+        validate_exists(form, "user_id"),
         validate_exists(form, "start_date")
         and validate_date(form["start_date"]),
-        validate_exists(form, "end_date") and validate_date(form["end_date"])
+        validate_exists(form, "end_date") and validate_date(form["end_date"]),
```
calebsander: Don't need trailing comma; not sure if intended

donut/templates/campus_positions.html
```diff
@@ -93,3 +135,22 @@ <h1 id="pagetitle" > The Ascit Position Service </h1>
     {{all_positions|safe}});
 </script>
 {% endblock %}
+{% block scripts %}
+    {{ super()  }}
```
calebsander: Nit: more spaces on right side of `super()` than left

calebsander (changes requested):

calebsander (approved):

PR #83, opened 2018-10-24
Fix get_position_holders and get_positions_held

davistardif:
### Summary
- Fix query and tests for get_position_holders
- Change get_positions_held to account for position relations

### Test Plan
- make test

### DIFF
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -77,20 +77,26 @@ def get_position_holders(pos_id):
     query += """FROM positions p LEFT JOIN position_relations pr
              ON p.pos_id=pr.pos_id_to INNER JOIN position_holders ph
              ON ph.pos_id=p.pos_id OR pr.pos_id_from=ph.pos_id
-             NATURAL JOIN members WHERE p.pos_id in (%s) OR pr.pos_id_to in (%s)""" % (
-        format_string, format_string)
+             NATURAL JOIN members WHERE p.pos_id in (%s)""" % (format_string)
 
     with flask.g.pymysql_db.cursor() as cursor:
-        cursor.execute(query, pos_id * 2)
+        cursor.execute(query, pos_id)
         return cursor.fetchall()
 
 
 def get_positions_held(user_id):
-    ''' Returns a list of all position id's held (directly) 
+    ''' Returns a list of all position id's held (directly or indirectly) 
     by the given user. If no positions are found, [] is returned. '''
-    query = 'SELECT pos_id FROM position_holders WHERE user_id = %s'
+    query = '''SELECT pos_id FROM position_holders ph 
+               WHERE user_id = %s
+               UNION 
+               SELECT pos_id FROM (
+               SELECT pos_id_to as pos_id FROM position_holders ph 
+               JOIN position_relations pr
+               ON ph.pos_id = pr.pos_id_from WHERE user_id = %s
+               ) sub'''
     with flask.g.pymysql_db.cursor() as cursor:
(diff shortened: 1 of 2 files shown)


tests/modules/groups/test_groups.py
```diff
@@ -54,23 +54,25 @@ def test_get_group_positions_data(client):
 def test_get_position_holders(client):
     res = helpers.get_position_holders(5)
     assert len(res) == 2
-    assert res[0]["first_name"] == "Sean"
-    assert res[1]["first_name"] == "Robert"
+    res = [res[0]['first_name'], res[1]['first_name']]
+    assert 'Robert' in res and 'Sean' in res
```
calebsander: Could `assert set(row['first_name'] for row in res) == set(['Robert', 'Sean'])`. Ditto for others.

donut/modules/groups/helpers.py
```diff
@@ -77,20 +77,26 @@ def get_position_holders(pos_id):
...
+    ''' Returns a list of all position id's held (directly or indirectly) 
     by the given user. If no positions are found, [] is returned. '''
-    query = 'SELECT pos_id FROM position_holders WHERE user_id = %s'
+    query = '''SELECT pos_id FROM position_holders ph 
+               WHERE user_id = %s
+               UNION 
+               SELECT pos_id FROM (
+               SELECT pos_id_to as pos_id FROM position_holders ph 
+               JOIN position_relations pr
+               ON ph.pos_id = pr.pos_id_from WHERE user_id = %s
```
calebsander, line 96: I think you might be able to just add `WHERE user_id = %s` around the whole `UNION` clause, so you don't need it in 2 places

calebsander (dismissed):

tests/modules/groups/test_groups.py
```diff
@@ -54,23 +54,25 @@ def test_get_group_positions_data(client):
...
+    assert 'David' in res and 'Sean' in res and 'Robert' in res
 
 
 def test_get_positions_held(client):
     res = helpers.get_positions_held(4)
     assert len(res) == 2
     assert 4 in res and 5 in res
+    res = helpers.get_positions_held(2)
+    assert len(res) == 3
+    assert 1 in res and 4 in res and 5 in res
```
majesticfish, line 76: Would be nice if there was another case where you tested user_id=1 since in both of the cases that you test, the user has a position through position relations (whereas user_id only holds a single position i think)

tests/modules/groups/test_groups.py
```diff
@@ -54,23 +54,25 @@ def test_get_group_positions_data(client):
...
-    assert res[1]["first_name"] == "Robert"
+    res = [res[0]['first_name'], res[1]['first_name']]
+    assert 'David' in res and 'Robert' in res
     res = helpers.get_position_holders([1, 5])
     assert len(res) == 3
-    assert res[0]["first_name"] == "David"
-    assert res[1]["first_name"] == "Robert"
-    assert res[2]["first_name"] == "Sean"
+    res = [res[0]['first_name'], res[1]['first_name'], res[2]['first_name']]
+    assert 'David' in res and 'Sean' in res and 'Robert' in res
```
majesticfish: nit: since things are getting a little messy, it'd be good to separate different tests cases with a newline. Don't worry about refactoring the whole file though, just this method is fine

majesticfish (dismissed):
Mostly good, just two minor things

tests/modules/groups/test_groups.py
davistardif: Thanks, will use going forward

PR #85, opened 2018-10-24
Update for Flask vulnerability.

joshc:
### Summary
- Updates to Makefile and requirements.txt

### Test Plan
- Run make update-packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: python
 python:
   - "3.6"
 services: mysql
-env:
-  - FLASK=0.10.1
 before_install:
     sudo apt-get install libmariadbclient-dev
 install: "pip install -r requirements.txt"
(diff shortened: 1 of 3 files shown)


calebsander:
Looks like there's a bug in 0.12.3 that's causing the tests to fail: https://github.com/dpgaspar/Flask-AppBuilder/issues/741
Sounds like 0.12.4 has this fixed

calebsander:
Can we remove the `FLASK` environment variable from `.travis.yml`? I don't think it's doing anything and it's out of sync:
````yaml
env:
  - FLASK=0.10.1
````

joshc:
@calebsander FYI, updating to the latest version of Flask (1.0.2) yields errors in some tests regarding voting. We can stick with this current version for now, but unsure why errors come up. Perhaps you can look into it?

calebsander (approved):

PR #88, opened 2018-11-07
Created Add Group and Delete Group helpers functions

majesticfish:
### Summary
- Two new functions in modules/groups/helpers.py
- helpers.create_group() creates a new group with specified attributes, also automatically creates a position associated with the group called "Member"
- helpers.delete_group() removes a group and all its positions
### Test Plan
- Two new test cases for these methods

### DIFF
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -242,6 +242,7 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(s, (pos_id, user_id, start_date, end_date))
 
+
 def create_group(group_name,
                  group_desc="",
                  group_type="",
@@ -273,7 +274,8 @@ def create_group(group_name,
         cursor.execute(query, (group_name, group_desc, group_type, newsgroups,
                                anyone_can_send, members_can_send, visible,
                                admin_control_members))
-        add_position(cursor.lastrowid, "Member");
+        add_position(cursor.lastrowid, "Member")
+
 
 def delete_group(group_id):
     '''
@@ -290,6 +292,7 @@ def delete_group(group_id):
     s = s % group_id
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(s)
+
 
 def get_members_by_group(group_id):
     '''
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -279,6 +279,7 @@ def create_group(group_name,
     add_position(new_group_id, "Member")
     return new_group_id
 
+
 def delete_group(group_id):
     '''
     Deletes the group specified with the group_id and all assocaited
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -278,6 +278,7 @@ def create_group(group_name,
     add_position(new_group_id, "Member")
     return new_group_id
 
+
 def delete_group(group_id):
     '''
     Deletes the group specified with the group_id and all assocaited
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -63,7 +63,7 @@ def get_position_holders(pos_id):
     and Y is linked to X
 
     Arguments:
-        pos_id:     The position to look up -- may be a single int or a list of int's 
+        pos_id:     The position to look up -- may be a single int or a list of int's
 
     Returns:
         results:    A list where each element describes a user who holds the
@@ -85,13 +85,13 @@ def get_position_holders(pos_id):
 
 
 def get_positions_held(user_id):
-    ''' Returns a list of all position id's held (directly or indirectly) 
+    ''' Returns a list of all position id's held (directly or indirectly)
     by the given user. If no positions are found, [] is returned. '''
-    query = '''SELECT pos_id FROM position_holders ph 
+    query = '''SELECT pos_id FROM position_holders ph
                WHERE user_id = %s
-               UNION 
+               UNION
                SELECT pos_id FROM (
-               SELECT pos_id_to as pos_id FROM position_holders ph 
+               SELECT pos_id_to as pos_id FROM position_holders ph
                JOIN position_relations pr
                ON ph.pos_id = pr.pos_id_from WHERE user_id = %s
                ) sub'''
@@ -102,7 +102,7 @@ def get_positions_held(user_id):
 
 
 def get_position_id(group_name, position_name):
-    ''' Returns the position id associated with the given group name and 
+    ''' Returns the position id associated with the given group name and
     position name '''
     query = '''SELECT pos_id FROM positions WHERE pos_name = %s
     AND group_id = (SELECT min(group_id) FROM groups WHERE group_name = %s)'''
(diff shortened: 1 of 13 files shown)


donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,63 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+    position holders entries and positions
+
+    Arguments:
+        group_id: id of the group to be deleted
+    '''
+    group_positions = get_group_positions(group_id)
+    for pos in group_positions:
+        delete_position(pos["pos_id"])
+    s = "DELETE FROM groups WHERE group_id=%d"
+    s = s % group_id
```
davistardif: Why do this here instead of at the `cursor.execute()` call?

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,63 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+        "members_can_send", "visible", "admin_control_members"
+    ]
+    query = "INSERT INTO groups (" + ','.join(fields) + ")"\
+        "VALUES (%s, %s, %s, %r, %r, %r, %r, %r)"
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (group_name, group_desc, group_type, newsgroups,
+                               anyone_can_send, members_can_send, visible,
+                               admin_control_members))
+        add_position(cursor.lastrowid, "Member")
+
```
davistardif, line 277: Would be nice to return the group_id for the group that just got created. For example, you could use the returned group_id in place of a constant in your tests, so if someone adds a new group to the test data it doesn't break your tests

davistardif (commented):
Some small suggestions. Other than that, looks good to me

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,63 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+        memberse_can_send: Toggles if members can send emails to this group
+        visible: Toggles if the group is visible
+        admin_control_members: toggles if administrator can control membership
+    """
+    fields = [
+        "group_name", "group_desc", "type", "newsgroups", "anyone_can_send",
+        "members_can_send", "visible", "admin_control_members"
+    ]
+    query = "INSERT INTO groups (" + ','.join(fields) + ")"\
+        "VALUES (%s, %s, %s, %r, %r, %r, %r, %r)"
```
calebsander: Is there a reason why these are not all `%s`?

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,63 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+    '''
+    Deletes the group specified with the group_id and all assocaited
+    position holders entries and positions
+
+    Arguments:
+        group_id: id of the group to be deleted
+    '''
+    group_positions = get_group_positions(group_id)
+    for pos in group_positions:
+        delete_position(pos["pos_id"])
```
calebsander: Couldn't this be accomplished with an `ON DELETE CASCADE` in the `positions` and `position_holders` tables?

calebsander: Couldn't this be accomplished in SQL instead with an `ON DELETE CASCADE`?

donut/modules/groups/helpers.py
calebsander: Why are these not all `%s`?

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,63 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+        "group_name", "group_desc", "type", "newsgroups", "anyone_can_send",
+        "members_can_send", "visible", "admin_control_members"
+    ]
+    query = "INSERT INTO groups (" + ','.join(fields) + ")"\
+        "VALUES (%s, %s, %s, %r, %r, %r, %r, %r)"
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (group_name, group_desc, group_type, newsgroups,
+                               anyone_can_send, members_can_send, visible,
+                               admin_control_members))
+        add_position(cursor.lastrowid, "Member")
```
calebsander: Put `cursor.lastrowid` in a local variable and call `add_position()` from outside the `with` block. No need to hold multiple cursors at the same time.

calebsander (changes requested):

donut/modules/groups/helpers.py
majesticfish: They are boolean values, does the formatting still work if it is `%s`?

calebsander: I would imagine so; `'%s' % True` and `'%r' % True` give the same value

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,61 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+        visible: Toggles if the group is visible
+        admin_control_members: toggles if administrator can control membership
+    """
+    fields = [
+        "group_name", "group_desc", "type", "newsgroups", "anyone_can_send",
+        "members_can_send", "visible", "admin_control_members"
+    ]
+    query = "INSERT INTO groups (" + ','.join(fields) + ")"\
+        "VALUES (%s, %s, %s, %s, %s, %s, %s, %s)"
+    new_group_id = -1
```
calebsander: Why initialize this value if it's always overwritten?

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,61 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+        anyone_can_send: Toggles if anyone can send emails to this group
+        memberse_can_send: Toggles if members can send emails to this group
+        visible: Toggles if the group is visible
+        admin_control_members: toggles if administrator can control membership
+    """
+    fields = [
+        "group_name", "group_desc", "type", "newsgroups", "anyone_can_send",
+        "members_can_send", "visible", "admin_control_members"
+    ]
+    query = "INSERT INTO groups (" + ','.join(fields) + ")"\
```
calebsander: If `fields` are fixed, might be cleaner to just hard-code them into `query`

tests/modules/groups/test_groups.py
```diff
@@ -159,6 +159,34 @@ def test_create_pos_holders(client):
...
+        "members_can_send", "newsgroups", "visible", "admin_control_members"
+    ]) == {
+        "group_name": "Page House",
+        "group_desc": "May the work be light",
+        "type": "House",
+        "newsgroups": 0,
+        "anyone_can_send": 0,
+        "members_can_send": 1,
+        "visible": 1,
+        "admin_control_members": 1,
```
calebsander: Seems a little counter-intuitive that these values are booleans when passed to `create_group()` but integers in the result of `get_group_data()`. Would be nice to convert them back to booleans.

donut/modules/groups/helpers.py
```diff
@@ -237,12 +237,61 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
...
+    '''
+    Deletes the group specified with the group_id and all assocaited
+    position holders entries and positions
+
+    Arguments:
+        group_id: id of the group to be deleted
+    '''
+    s = "DELETE FROM groups WHERE group_id=%s"
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(s, (group_id))
```
calebsander: Parens unnecessary

calebsander (approved):

PR #89, opened 2018-11-10
Update to Flask 1.0.x

calebsander:
Removed dependencies which Flask already bundles and fixed `voting` tests so they don't care whether response JSON is pretty-printed


joshc (approved):
nice. unsure what the standard is for specifying specific versions of dependencies, weird that pip doesn't handle it automatically. this looks good and fixes the tests. follow-up: some way to make sure that everyone's dev virtual envs are running the same stuff hah.

PR #90, opened 2018-11-21
Delete Position Holders Pill

majesticfish:
### Summary
- Added a delete position holders pill on the positions tab
- Added new 'get_direct_holders' helper function
- Added /1/positions/<int>/delete POST endpoint

### Test Plan
- Added test case for helper function
- Manual e2e test:
-- Add new user to hold a position (with the right time frame)
-- Reload page to see he has been added 
-- Navigate to remove position holder tab and remove him
-- See that he is no longer there

### DIFF
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -54,6 +54,7 @@ def get_group_positions(group_id):
         cursor.execute(query, [group_id])
         return list(cursor.fetchall())
 
+
 def get_direct_position_holders(pos_id):
     """
     Queries the database and returns a list of all members and their names
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -54,6 +54,29 @@ def get_group_positions(group_id):
         cursor.execute(query, [group_id])
         return list(cursor.fetchall())
 
+def get_direct_position_holders(pos_id):
+    """
+    Queries the database and returns a list of all members and their names
+    that currently hold the position specified by the pos_id. This only
+    includes people that are directly holding the position as specified
+    by the position_holders table
+
+    Arguments:
+        pos_id: the position id to be queried
+    Returns:
+        results:    A list where each element describes a user who holds the
+                    position. Each element is a dict with key:value of
+                    columnname:columnvalue
+    """
+    fields = ["user_id", "first_name", "last_name", "start_date", "end_date"]
+    query = "SELECT " + ', '.join(fields) + " "
+    query += """FROM position_holders NATURAL JOIN members
+                WHERE pos_id =%s"""
+
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [pos_id])
+        return cursor.fetchall()
+
 
 def get_position_holders(pos_id):
     """
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -69,14 +69,12 @@ def get_direct_position_holders(pos_id):
                     position. Each element is a dict with key:value of
                     columnname:columnvalue
     """
-    fields = ["user_id", "first_name", "last_name", "start_date", "end_date"]
-    query = "SELECT " + ', '.join(fields) + " "
-    query += """FROM position_holders NATURAL JOIN members
-                WHERE pos_id =%s"""
+    query = """SELECT user_id, first_name, last_name, start_date, end_date
+            FROM position_holders NATURAL JOIN members WHERE pos_id = %s"""
 
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, [pos_id])
-        return cursor.fetchall()
+        return list(cursor.fetchall())
 
 
 def get_position_holders(pos_id):
@@ -265,7 +263,6 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
     end_date) VALUES (%s, %s, %s, %s)"""
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(s, (pos_id, user_id, start_date, end_date))
-        cursor.close()
 
 
 def delete_position_holder(pos_id, user_id):
--- a/donut/modules/groups/helpers.py
+++ b/donut/modules/groups/helpers.py
@@ -63,7 +63,7 @@ def get_position_holders(pos_id):
     and Y is linked to X
 
     Arguments:
-        pos_id:     The position to look up -- may be a single int or a list of int's 
+        pos_id:     The position to look up -- may be a single int or a list of int's
 
     Returns:
         results:    A list where each element describes a user who holds the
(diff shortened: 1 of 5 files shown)


donut/modules/groups/helpers.py
```diff
@@ -55,6 +55,30 @@ def get_group_positions(group_id):
...
+
+    Arguments:
+        pos_id: the position id to be queried
+    Returns:
+        results:    A list where each element describes a user who holds the
+                    position. Each element is a dict with key:value of
+                    columnname:columnvalue
+    """
+    fields = ["user_id", "first_name", "last_name", "start_date", "end_date"]
+    query = "SELECT " + ', '.join(fields) + " "
```
calebsander: If `fields` are hard-coded, it seems like it would be cleaner to just write out the full query string (as a triple-quoted string)

donut/modules/groups/helpers.py
```diff
@@ -55,6 +55,30 @@ def get_group_positions(group_id):
...
+    query += """FROM position_holders NATURAL JOIN members
+                WHERE pos_id =%s"""
```
calebsander: Nit: asymmetric spacing around `=`

donut/modules/groups/helpers.py
```diff
@@ -237,10 +261,25 @@ def create_position_holder(pos_id, user_id, start_date, end_date):
         start_date: Starting date of the holding period, format is 'yyyy-mm-dd'
         end_date: end date of the hold period
     '''
-    s = """INSERT INTO position_holders (pos_id, user_id, start_date, 
+    s = """INSERT INTO position_holders (pos_id, user_id, start_date,
     end_date) VALUES (%s, %s, %s, %s)"""
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(s, (pos_id, user_id, start_date, end_date))
+        cursor.close()
```
calebsander: Should be unnecessary (here and below); when a `with` block exits, the context object is automatically closed

donut/static/js/groups/positions.js
```diff
@@ -26,6 +26,7 @@ function populateAdminGroups(approvedGroupIds, approvedGroupNames) {
         $('#groupCreate option:last').after(newOption);
         $('#groupDel option:last').after(newOption);
         $('#groupPosHold option:last').after(newOption);
+        $('#groupDelPosHold option:last').after(newOption);
```
calebsander, line 31: Hmm, maybe it's too late to change this style now; but it would probably be faster to select all the `<select>`s and just call `.append()`, rather than looking for the last `option` in each `select`. Could do something like this, where the class `group-select` is added to all the relevant `<select>`s:
```js
$('.group-select').append(newOption)
```

donut/static/js/groups/positions.js
```diff
@@ -151,6 +152,51 @@ function groupPosHoldChange() {
...
+    $('#posIdDelHold').find('option').remove();
+    $('#posIdDelHold').append('<option> Pick a position </option>');
+    populateListOfPositions(groupIndex, '#posIdDelHold');
+}
+
+/*
+ * Converts a dateString to a more readable format
+ */
+function formatDate(dateString) {
+    if (dateString == null) {
```
calebsander, line 172: Don't think the `dateString == null` case should be handled inside this function. Would make more sense for the caller to deal with this. Imagine what happens when `data[i].start_date` and `data[i].end_date` are `null`: you will see something like `(null to null)` as the option value. Are these dates even allowed to be `NULL` in the database?

donut/static/js/groups/positions.js
```diff
@@ -151,6 +152,51 @@ function groupPosHoldChange() {
...
+    var positionIndex = $('#posIdDelHold').val();
+    var url = '/1/positions/' + positionIndex + '/direct/';
+    $.ajax({
+        url: url,
+        success: function(data){
+            $('#delName').find('option').remove();
+            for (var i = 0; i < data.length; i++) {
+                var newOption = '<option value=' + data[i].user_id + '>' +
+                    data[i].first_name + ' ' + data[i].last_name +
+                    ' (' + formatDate(data[i].start_date) + " to " +
```
calebsander: Nit: double-quoted string

donut/static/js/groups/positions.js
```diff
@@ -214,14 +261,31 @@ $(document).ready(function() {
             }
         });
     });
+    $('#submitDelPosHoldBtn').click(function(e) {
+        e.preventDefault();
```
calebsander: Should be able to avoid calling `preventDefault()` if you add the attribute `type = 'button'` to the button, so it doesn't act as a submission button

donut/templates/campus_positions.html
```diff
@@ -127,6 +130,36 @@ <h2 id="pagetitle">The Ascit Position Service</h2>
                         </button>
                     </form>
                 </div>
+                <div id="deleteHold" class="tab-pane fade in">
+                    <form id="deleteHoldForm">
+                        <div class="form-group">
+                            <label class="control-label col-sm-2" for="groupDelPosHold">
+                                Group Name:</label>
```
calebsander: Would be easier to read if closing `</label>` were aligned with `<label>` (and below too)

tests/modules/groups/test_groups.py
```diff
@@ -51,6 +51,16 @@ def test_get_group_positions_data(client):
     assert helpers.get_group_positions(4) == []
 
 
+def test_get_direct_position_holders(client):
+    res = helpers.get_direct_position_holders(5)
+    assert len(res) == 1
+    assert set(row['first_name'] for row in res) == set(['Sean'])
```
calebsander, line 56: If this were a `list` comparison, wouldn't have to check the length

donut/modules/groups/helpers.py
```diff
@@ -55,6 +55,30 @@ def get_group_positions(group_id):
...
+                    columnname:columnvalue
+    """
+    fields = ["user_id", "first_name", "last_name", "start_date", "end_date"]
+    query = "SELECT " + ', '.join(fields) + " "
+    query += """FROM position_holders NATURAL JOIN members
+                WHERE pos_id =%s"""
+
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [pos_id])
+        return cursor.fetchall()
```
calebsander: Should probably convert to `list` for consistency

donut/static/js/groups/positions.js
majesticfish: I think this makes sense, maybe I can put this in particular as a TODO and have some of the new members look into as a starter task?

donut/static/js/groups/positions.js
majesticfish: The issue is that in the permission holders table, null date values are allowed. 

In the future when the Current_Position_holders View comes online we won't run into this issue. Since we're still pulling from data that could have null start/end date (and we still might want to see those positions), i think it makes sense to have it here for now.

donut/static/js/groups/positions.js
majesticfish: Ok, it makes sense to do this for the other buttons as well

donut/static/js/groups/positions.js
calebsander: Sure, sounds reasonable

calebsander:
Superseded by #142

PR #91, opened 2018-11-21
Created a new view for current positions

majesticfish:
Work in progress
## Summary
- Added view that only selects position holders that have not expired
### Test Plan
- Need to add a test case to verify that this actually works
### Refactor
- For every place that queries position_holder, we need to decide if we should change it to current position holder or not
- Should we force every position_holder element to have a start date and end date? What sort of refactoring would that entail?

### DIFF
--- a/sql/donut.sql
+++ b/sql/donut.sql
@@ -4,6 +4,7 @@
 DROP TABLE IF EXISTS position_relations;
 DROP VIEW IF EXISTS group_house_membership;
 DROP VIEW IF EXISTS group_houses;
+DROP VIEW IF EXISTS curren_position_holders;
 DROP TABLE IF EXISTS position_holders;
 DROP TABLE IF EXISTS positions;
 DROP TABLE IF EXISTS groups;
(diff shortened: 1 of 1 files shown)


sql/donut.sql
```diff
@@ -4,6 +4,7 @@
 DROP TABLE IF EXISTS position_relations;
 DROP VIEW IF EXISTS group_house_membership;
 DROP VIEW IF EXISTS group_houses;
+DROP VIEW IF EXISTS curren_position_holders;
```
calebsander, line 7: Missing `t`

sql/donut.sql
```diff
@@ -178,6 +179,14 @@ CREATE VIEW group_house_membership AS (
         OR UPPER(pos_name) = UPPER('Social Member')
 );
 
+-- Current Position Holders View
+-- This view only contains position holders that are currently valid
+CREATE VIEW current_position_holders AS (
+    SELECT hold_id, pos_id, user_id, start_date, end_date
+    FROM position_holders 
+    WHERE CURDATE() between start_date and end_date
```
calebsander, line 187: Nit: `between` and `and` should be upper-cased for consistency

calebsander:
Superseded by #142

PR #92, opened 2018-12-06
Scheduler and planner

calebsander:
### Summary
- Added planner and scheduler
- Added import script for registrar files: `donut/modules/courses/utils/import_registrar.py`

### Test Plan
- Run tests
- Try interface of scheduler and planner (searching, adding/dropping, refreshing page, switching terms on scheduler, checking core requirements on planner)


sql/courses.sql
```diff
@@ -0,0 +1,67 @@
...
+    PRIMARY KEY (grades_type_id)
+);
+
+CREATE TABLE sections (
+    course_id       INT           NOT NULL,
+    section_number  TINYINT       NOT NULL, -- e.g. 1
+    instructor_id   INT           NOT NULL,
+    grades_type_id  INT           NOT NULL,
+    times           VARCHAR(100)  NOT NULL, -- e.g. MWF 13:00 - 13:55
+    locations       VARCHAR(100)  NOT NULL, -- e.g. 213 ANB; currently unused
```
davistardif: If `locations` is unused, why not allow `NULL`?

calebsander: The upload script always puts the data in there, it just isn't used by the scheduler or planner. I figured we might as well store this data since it is included in the registrar's file.

davistardif:
Links to scheduler and planner are missing from the navbar
Drop them into `donut/templates/includes/header.html`

calebsander:
I already did that, check the diff: https://github.com/ASCIT/donut/pull/92/files#diff-a45d40180b13503783eda366636ff6c4

davistardif:
You're right, I checked out an outdated version of your branch. Oops

davistardif:
Drag and drop on the planner isn't working for me in Firefox
Might be related to [this](http://mereskin.github.io/dnd/)?

calebsander:
Good catch, I think I fixed it. Can you try it in Firefox again?

davistardif:
Fixed!

sql/courses.sql
joshc: can probably surface this in scheduler, i personally would find it useful

donut/static/js/courses/search.js
```diff
@@ -0,0 +1,34 @@
...
+	' ': true,
+	'/': true,
+	undefined: true
+}
+
+/**
+ * Filters a list of courses by whether they match a query string.
+ * Sorts courses from highest matching score to lowest.
+ */
+function searchCourses(courses, query) {
```
joshc, line 11: did you come up with this algorithm lol. a point to consider would be to have results like ma1 appear at the top if someone types ma1, and not ma10* since those appear before ma1a. some weighting of how short the string following the thing is, or ordering alphabetic characters before numeric characters if they come after the course number. how fun! implement literally for the laziest of lazies or for fun :D

donut/modules/courses/utils/import_registrar.py
```diff
@@ -0,0 +1,139 @@
...
+import re
+import sys
+from donut.pymysql_connection import make_db
+from donut.modules.courses import helpers
+
+COURSE_MATCH = r'^(?P<dept>[A-Za-z/]+) *0*(?P<number>[0-9]+[A-Z]*)$'
+
+db = make_db('dev')
+
+_, filename, year, term = sys.argv
```
joshc: should print usage statement if incorrect args provided. look into argparse library for autogeneration of usage statements and easy specification of command line arguments

donut/modules/courses/utils/import_registrar.py
```diff
@@ -0,0 +1,139 @@
...
+        row_data = dict(zip(headers, row))
+        course = row_data['Course Number']
+        if course not in courses:
+            try:
+                match = re.search(COURSE_MATCH, course)
+                dept = match.group('dept')
+                number = match.group('number').lower()
+                units = row_data['Units']
+                if units == '+': units = '0-0-0'
+                lecture, lab, homework = map(float, units.split('-'))
```
joshc, line 75: does this mean there are non-integer unit splits some times :O

donut/modules/courses/utils/import_registrar.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
```
joshc, line 1: in general should comment this script a little more. specifically, consider the case where the format of data registrar gives us changes. then there could be changes that need to be made here. also, might be good to have a README in general detailing places to change in event of new/removed major, new/removed column, etc.

donut/modules/courses/utils/import_registrar.py
```diff
@@ -0,0 +1,139 @@
...
+
+# Example command: ./import_registrar.py donut_courses_FA2018-19_180515.txt 2018 FA
```
joshc: potentially fix the directory where the script looks for new course files to keep everything centralized and not split among multiple user directories. couple of ideas to toss around, or also to toss away: have whoever is handling this throw the data in a directory and then run this script (which can be modified to just use the most recently added file, would require file name to be standardized), run a script to upload the file path / text file to the database along with the term / year, and periodically clean out, only keeping the last 4 years (unsure if you want to drop courses from the database that are removed in new years). should keep the amount of course source files down. also course source lol

donut/modules/courses/helpers.py
```diff
@@ -0,0 +1,247 @@
...
+        sections.append({
+            'number': section['section_number'],
+            'instructor': section['instructor'],
+            'grades': section['grades_type'],
+            'times': section['times']
+        })
+    courses = course_sections.values()
+    for course in courses:
+        course['sections'].sort(key=lambda section: section['number'])
+    return sorted(courses, key=lambda course: course['number'].lower())
```
joshc, line 199: why would .lower() be necessary here? shouldn't they be stored in lowered format.

donut/modules/courses/routes.py
```diff
@@ -0,0 +1,113 @@
+import flask
+import pymysql
+
+from donut.modules.courses import blueprint, helpers
+
+YEARS = {1: 'Freshman', 2: 'Sophomore', 3: 'Junior', 4: 'Senior'}
+WEEK_DAYS = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday']
+SCHEDULER_START_HOUR = 8  # 8 AM
+SCHEDULER_END_HOUR = 23  # 11 PM
```
joshc, line 9: are there courses this late?

donut/modules/courses/templates/planner.html
```diff
@@ -0,0 +1,431 @@
...
+          menu = false,
+          Ch3 = false, labUnits = 0,
+          PE = 0,
+          hums = 0
+      termLists.each(function(_, termList) {
+        $(termList).find('.list-group-item').each(function(_, course) {
+          var courseName = $(course).attr('course'),
+            units = Number($(course).attr('units'))
+          totalUnits += units
+          switch (courseName) {
```
joshc, line 368: lmao, better hope this doesn't change

donut/modules/courses/templates/planner.html
```diff
@@ -0,0 +1,431 @@
...
+      margin-bottom: 10px;
+    }
+  </style>
+{% endblock %}
+{% block scripts %}
+  {{ super() }}
+  <script src='{{ url_for("static", filename="js/courses/search.js") }}'></script>
+  <script>
+    var SHOW_COURSES = 50
+    var HSS_DEPARTMENTS = {
```
joshc, line 176: use a set? idk what js has lol

donut/modules/courses/templates/planner.html
```diff
@@ -0,0 +1,431 @@
+{% extends 'layout.html' %}
```
joshc, line 1: general comments:

- is it possible to get the trash buttons to line up?
- legacy has these colorized, though i suppose it doesn't really matter lol
- would be useful to have a unit count per term
- when i try to add the same class to the same term twice, it gives me an error, but still shows it on the front end (doesn't log it on the back end as expected)
- having multiple planners for an user open seems a little strange, but i think it's fine actually.
- should the width of columns be fixed or something? the grid doesn't seem like it is mobile optimized, and i think it'll try to automatically resize itself when on mobile... maybe display less search results?
- if it splits line, should it pad again the border instead of pressing up? image

donut/modules/courses/templates/scheduler.html
```diff
@@ -0,0 +1,424 @@
+{% extends 'layout.html' %}
```
joshc, line 1: General comments:
- The check box for the sections are spaced and misaligned from the section number.
- Is it possible to make it so when you click in the box the check box is checked/unchecked? Also, for the pop up for the sections, if you uncheck the check box, the pop up disappears. I think it'd be better if it stayed, especially if there's no way to make the box check/uncheck the check box.
- The last hour is a cut off (or i guess it shouldn't be shown).
- When beginning a search, you probably want to reset the scroll bar to the top.
- Again, mobile behavior is somewhat strange (I just made the window very small, and you can see).
- If you open scheduler in two different tabs, and they both try to add the same course, the error that pops up on the second one is "Failed to drop course..."

donut/modules/courses/utils/import_registrar.py
joshc: Maybe also document procedure for loading in courses in wiki. Also, thinking about the above, maybe we wouldn't want to clean out the databases, but definitely the source files? for the former, it would probably be useful for additional features like ratemyprof, course recommender, etc.

joshc (changes requested):
this is probably not relevant to this feature specifically, but when you are on a page, and then you click log in, after you log in you normally want to return to the page but now it takes you to home screen.

donut/static/js/courses/search.js
calebsander: Yup, made this using a bunch of trial and error. Ma 1a will come up first if you type `ma 1a` but yeah, I can definitely change that

donut/modules/courses/utils/import_registrar.py
calebsander: Yes, SS/Psy 283B is 1.5-0-1.5 this coming term

donut/modules/courses/utils/import_registrar.py
calebsander: One of the goals of the way I designed the database was to make it so you never have to remove old courses. In theory, you should be able to throw away the courses file as soon as it's imported. I didn't want to force people to put the files in any specific place, so I figured it made sense for the user to always give a full path to the file.

donut/modules/courses/helpers.py
calebsander: `number` is perhaps not a great name for it, but that value includes the department, e.g. `CS 21`. The lower-casing is for sorting the departments (e.g. `Ch` should come before `CS` even though the ASCII code for `h` is higher)

donut/modules/courses/routes.py
calebsander: Ya, I'm taking a power hum next term that goes until 10:30

donut/modules/courses/templates/planner.html
calebsander: lol, at least it would be easy to remove it if the requirements change substantially

donut/modules/courses/templates/planner.html
calebsander: JS only added `Set` ~3 years ago so it's still not considered standard. Using a dictionary isn't too bad

calebsander:
I think I've addressed all your comments. Would be great if you could take another look whenever works for you.

joshc (dismissed):
- stylistically speaking, for the scheduler, it would be good to indicate what you should press (like add a box around the section if there is only one section)
image I was somewhat confused at first what I should be pressing to add a course to the schedule, and maybe also give this box and or the text and the text in the section an effect if you hover over it. OR (i'm not sure how hard this is to do) for the cases where there is only one section, make the entire box clickable.
- also does adding a border around the scheduler table make it look better? up to you :)
- holy moly didn't know that the legacy course planner also had course requirements lol

other than those i think it looks good 💯

PR #94, opened 2018-12-06
Avoid comparing numbers to strings

calebsander:
Fixed these warnings in the test output, which were due to `members.graduation_year` being compared to `''`:
```
tests/modules/directory_search/test_directory_search.py::test_value_lists
  /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pymysql/cursors.py:323: Warning: (1292, "Truncated incorrect DOUBLE value: ''")
    self._do_get_result()
tests/modules/directory_search/test_directory_search.py::test_search_page
  /home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pymysql/cursors.py:323: Warning: (1292, "Truncated incorrect DOUBLE value: ''")
    self._do_get_result()
```

### DIFF
--- a/donut/modules/directory_search/helpers.py
+++ b/donut/modules/directory_search/helpers.py
@@ -164,8 +164,11 @@ def execute_search(**kwargs):
         return cursor.fetchall()
 
 
-def members_unique_values(field):
-    query = 'SELECT DISTINCT ' + field + ' FROM members WHERE ' + field + ' IS NOT NULL AND ' + field + '!= "" ORDER BY ' + field
+def members_unique_values(field, string):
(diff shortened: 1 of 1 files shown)


joshc (approved):
Are these are the only two functions that use it? If so, I'm guessing you're building this with the intention that other things that use this will be built later down the line. If so, this is fine but add some comments to members_unique_values that says like this function should be used to extract unique attributes of members before merging.

calebsander:
Wasn't really intending for other things to use this in the future. I was mainly aiming to avoid duplicating code since the two functions do almost exactly the same thing.

PR #95, opened 2019-04-23
Masquerade

Thought-Weaver:
### Summary
- Added masquerading to auth.

### Test Plan
- Caleb, Ziyan, and myself all have permissions (admin) to test this.
- I've already added in several checks and tested numerous cases, but I'll update it if something is missing.


donut/modules/auth/routes.py
```diff
@@ -12,13 +13,52 @@ def login():
...
 
+    # Need to take care of the case when masquerading occurs.
+    username = None
+    mask = None
+
+    if mask_and_user is not None:
+        # Check for the character to split.
+        if ':' in mask_and_user:
+            # If it's there, we assume the second part is a mask name.
+            split = mask_and_user.split(':')
```
calebsander, line 27: Could just do the `split()` and then check how many elements are in the split

donut/modules/auth/routes.py
```diff
@@ -12,13 +13,52 @@ def login():
...
+
+            if len(split) > 2:
+                flask.flash('You cannot specify multiple masks at once.')
+                return flask.redirect(flask.url_for('auth.login'))
+
+            username, mask = split[0], split[1]
```
calebsander: `username, mask = split`

donut/modules/auth/routes.py
```diff
@@ -12,13 +13,52 @@ def login():
...
+                flask.flash('That person does not exist to mask.')
+                return flask.redirect(flask.url_for('auth.login'))
+        else:
+            username = mask_and_user
+
     if username is not None and password is not None:
         user_id = helpers.authenticate(username, password)
         if user_id is not None:
-            flask.session['username'] = username
+            permissions = auth_utils.get_permissions(username)
```
koivyr, line 40: Why do we need this?

donut/modules/auth/routes.py
```diff
@@ -12,13 +13,52 @@ def login():
...
+            if len(split) > 2:
+                flask.flash('You cannot specify multiple masks at once.')
+                return flask.redirect(flask.url_for('auth.login'))
+
+            username, mask = split[0], split[1]
+
+            # But check to make sure the mask exists.
+            if not auth_utils.get_user_id(mask):
+                flask.flash('That person does not exist to mask.')
+                return flask.redirect(flask.url_for('auth.login'))
```
calebsander: Are there any security implications of allowing unauthenticated requests to query whether a username exists? Seems like we should move this check to inside `if mask_perm` below, just to be safe.

calebsander (changes requested):

donut/modules/auth/routes.py
Thought-Weaver: In theory, masquerading should only be necessary for debugging from a particular user's perspective; ideally, we should be able to understand everything from their point of view, including permissions they currently have. If something needs to be modified, we have power ourselves to do that without needing to transfer admin privileges while masquerading.

donut/modules/auth/routes.py
calebsander: I actually meant instead of the `':' in mask_and_user` check; if the split contains only 1 element, then it's just the username

calebsander (approved):

PR #97, opened 2019-06-04
Calendar

koivyr:
### Summary
- Add calendar that can be shared to google calendars
### Test Plan
- Run tests
- Share a calendar to yourself, add something on google calendars, and see it pop up on donut and vice versa.


donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+from google_auth_oauthlib.flow import InstalledAppFlow
+from google.auth.transport.requests import Request
+from google.oauth2 import service_account
+from donut.modules.calendar.permissions import calendar_permissions
+from donut import auth_utils
+import json
+
+# DEBUGGING
+import time
+current_milli_time = lambda: int(round(time.time() * 1000))
```
calebsander: Are these still needed?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    'Blacker': calendar_permissions.BLACKER,
+    'Dabney': calendar_permissions.DABNEY,
+    'Fleming': calendar_permissions.FLEMING,
+    'Lloyd': calendar_permissions.LLOYD,
+    'Page': calendar_permissions.PAGE,
+    'Ricketts': calendar_permissions.RICKETTS,
+    'Ruddock': calendar_permissions.RUDDOCK,
+    'Other': calendar_permissions.OTHER,  # For clubs
+    'Athletics': calendar_permissions.
+    ATHLETICS  # For all ath team and sports related events?
```
calebsander: This line is a little hard to read; could you move the comment to above the line so yapf doesn't split it?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+        'dateTime': db_event['begin_time'].isoformat('T') + 'Z'
+    }
+    db_event['end'] = {'dateTime': db_event['end_time'].isoformat('T') + 'Z'}
+    db_event['organizer'] = {'displayName': db_event['calendar_tag']}
+
+
+def get_events_backup(begin_month=datetime.datetime.now().month,
+                      begin_year=datetime.datetime.now().year,
+                      end_month=datetime.datetime.now().month,
+                      end_year=datetime.datetime.now().year + 2):
```
calebsander: I think it might be better not to have default values, so the caller has to specify the range

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+            WHERE %s <= end_time AND begin_time <= %s
+            ORDER BY begin_time
+    """
+
+    min_time = datetime.datetime(int(begin_year), begin_month,
+                                 1).isoformat('T') + 'Z'
+    max_time = datetime.datetime(
+        int(end_year),
+        int(end_month), calendar.monthrange(
+            int(end_year), int(end_month))[1]).isoformat('T') + 'Z'
```
calebsander, line 88: Is converting these dates to strings necessary? Can they just be passed in as `datetime` objects?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def insert_event_to_db(calendar_tag,
+                       google_event_id,
+                       event_summary,
+                       event_description,
+                       event_location,
+                       begin_time,
+                       end_time,
+                       username=''):
```
calebsander: Maybe default to `None`?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+def set_event_dict(db_event):
+    db_event['id'] = db_event['google_event_id']
+    db_event['start'] = {
+        'dateTime': db_event['begin_time'].isoformat('T') + 'Z'
+    }
+    db_event['end'] = {'dateTime': db_event['end_time'].isoformat('T') + 'Z'}
+    db_event['organizer'] = {'displayName': db_event['calendar_tag']}
+
+
+def get_events_backup(begin_month=datetime.datetime.now().month,
```
calebsander, line 62: Would it make sense to consolidate this function with `get_all_events_backup()`? Could make the begin/end parameters optional

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    'Athletics': 'g2itc2p2r9affcc77l1d2vg47s@group.calendar.google.com'
+}
+
+
+def get_all_events_backup():
+    '''
+    Returns a list of all events from our db. 
+    '''
+    query = '''
+        SELECT * FROM calendar_events
```
calebsander: `ORDER BY begin_time`?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                db_events = cursor.fetchall()
+        else:
+            query = """
+                    SELECT * FROM calendar_events 
+                    WHERE %s <= end_time AND begin_time <= %s
+                    ORDER BY begin_time
+                """
+            with flask.g.pymysql_db.cursor() as cursor:
+                cursor.execute(query, [min_time, max_time])
+                db_events = cursor.fetchall()
```
calebsander: Why not just call `get_events_backup()`?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                'end']['date']
+            insert_event_to_db(calendar_tag, google_event_id, event_summary,
+                               event_description, event_location, begin_time,
+                               end_time)
+            to_be_deleted = ''
+
+            # Find the current google event in our db and make
+            # sure that we mark it as visited.
+            for db_event_key in db_events:
+                if db_event_key['google_event_id'] == google_event_id:
```
calebsander: This algorithm would be O(n) instead of O(n^2) if you constructed a `dict` mapping `google_event_id`s to events

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                    to_be_deleted = db_event_key
+                    db_events.remove(to_be_deleted)
```
calebsander: Deletion in a list is also quite slow, especially because `remove()` has to traverse the list to find the index of the element to delete. Would be better with a `dict`.

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+            event_description = i.get('description', '')
+            event_location = i.get('location', '')
+            begin_time = i['start']['dateTime'] if 'dateTime' in i[
+                'start'] else i['start']['date']
+            end_time = i['end']['dateTime'] if 'dateTime' in i['end'] else i[
+                'end']['date']
+            insert_event_to_db(calendar_tag, google_event_id, event_summary,
+                               event_description, event_location, begin_time,
+                               end_time)
+            to_be_deleted = ''
```
calebsander: Don't think this variable is actually needed below

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    max_time = datetime.datetime(
+        int(end_year),
+        int(end_month), calendar.monthrange(
+            int(end_year), int(end_month))[1]).isoformat('T') + 'Z'
+    try:
+
+        all_events = get_all_events() if all_data else get_events(
+            begin_month, begin_year, end_month, end_year)
+        if all_events == []:
+            return
```
calebsander: What is this special case for?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+            for db_event_key in db_events:
+                if db_event_key['google_event_id'] == google_event_id:
+                    to_be_deleted = db_event_key
+                    db_events.remove(to_be_deleted)
+                    break
+
+        for db_event_key in db_events:
+            delete_event_from_db(db_event_key['google_event_id'])
+        return all_events
+    except Exception as e:
```
calebsander, line 212: Is there a specific exception you're expecting? It seems a little dangerous to mask *all* exceptions

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                    calendarId=value,
+                    timeMin=min_time,
+                    timeMax=max_time,
+                    singleEvents=True,
+                    orderBy='startTime').execute()
+                events = events_result.get('items', [])
+                all_events.extend(events)
+                next_page = 'nextPageToken' in events_result
+        return all_events
+    except:
```
calebsander: Same question—silently catching all exceptions is a little dangerous

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+        min_time = datetime.datetime(int(begin_year), begin_month,
+                                     1).isoformat('T') + 'Z'
+        max_time = datetime.datetime(
+            int(end_year),
+            int(end_month),
+            calendar.monthrange(int(end_year),
+                                int(end_month))[1]).isoformat('T') + 'Z'
+        all_events = []
+
+        for key, value in cal_id.items():
```
calebsander: `key` seems unused, so I think this could be changed to `for account in cal_id.values()`

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                    orderBy='startTime').execute()
+                events = events_result.get('items', [])
+                all_events.extend(events)
+                next_page = 'nextPageToken' in events_result
+        return all_events
+    except:
+        return []
+
+
+def get_all_events():
```
calebsander: Can this be combined with `get_events()`?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+            }
+            if get_permission()[cal_tag]:
+                created_event = service.events().insert(
+                    calendarId=cal_id[cal_tag], body=event).execute()
+                insert_event_to_db(cal_tag, created_event['id'], name,
+                                   description, location, begin_time, end_time,
+                                   flask.session['username'])
+        return ''
+    except Exception as e:
+        return e.get('message', e)
```
calebsander: Might be good to log the stacktrace as well

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+        a single value for updates
+    begin_time: in format YYYY-MM-DDThh:mm:00Z
+    end_time: as above. 
+    update: bool for it we are adding events or updating them
+    location: not required; location for event
+    """
+    try:
+        if update:
+            event = service.events().get(
+                calendarId=cal_id[calendar_tag[0]],
```
calebsander: Maybe store `calendar_tag[0]` in a variable since it is used several times?

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    ''' 
+    Returns permissions list related to calendars 
+    '''
+    perms = {}
+    # If they have edit permissions on anything, the events pages and stuff
+    # will show up.
+    anyy = False
+    for tag in cal_permissions:
+        perms[
+            tag] = 'username' in flask.session and auth_utils.check_permission(
```
calebsander: Should check `'username' in flask.session` and compute `flask.session['username']` outside the loop

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                flask.session['username'], cal_permissions[tag])
```
calebsander: Can iterate over `cal_permissions.items()` instead

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    try:
+        if update:
+            event = service.events().get(
+                calendarId=cal_id[calendar_tag[0]],
+                eventId=event_id).execute()
+            event['start']['dateTime'] = begin_time
+            event['end']['dateTime'] = end_time
+            event['start']['date'] = None
+            event['end']['date'] = None
+            event['end']['timeZone'] = 'America/Los_Angeles'
```
calebsander: Should probably be a constant at the top of the file

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+                    'timeZone': 'America/Los_Angeles',
+                },
+                'end': {
+                    'dateTime': end_time,
+                    'timeZone': 'America/Los_Angeles',
+                },
+                'location': location,
+                'visibility': 'public',
+            }
+            if get_permission()[cal_tag]:
```
calebsander: Should move the call to `get_permission()` outside the loop

donut/modules/calendar/helpers.py
```diff
@@ -0,0 +1,412 @@
...
+    except Exception as e:
+        return e.get('message', e)
+
+
+def share_calendar(calendars, email, permission_level):
+    '''
+    Shares a set of calendars to a given email
+    '''
+    try:
+        for index, cal_name in enumerate(calendars):
```
calebsander: Is `index` needed?

donut/modules/calendar/routes.py
```diff
@@ -0,0 +1,142 @@
...
+    end_date = flask.request.form.get('end_date')
+    end_hour = flask.request.form.get('end_hour')
+    end_time = flask.request.form.get('end_minute')
+    begin_datetime = datetime.datetime(
+        int(start_date[0:4]),
+        int(start_date[5:7]),
+        int(start_date[8:10]), int(start_hour), int(start_time))
+    end_datetime = datetime.datetime(
+        int(end_date[0:4]),
+        int(end_date[5:7]), int(end_date[8:10]), int(end_hour), int(end_time))
```
calebsander: This is a little gross; could use something like `start_year, start_month, start_date = map(int, start_date.split('-'))`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+  'Ruddock':'#74deff',
+  'Ricketts':'#b974ff',
+  'Athletics':'#ff0000',
+  'Other':'#868686',
+
+};
+function isFuture(curEvent)
+{
+
+    var timeStr = curEvent['end']['dateTime'] || curEvent['end']['date'];
```
calebsander: Can use `curEvent.end.dateTime` and `curEvent.end.date` in JS

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+    var year = parseInt(timeStr.substring(0, 4), 10);
+    var day = parseInt(timeStr.substring(8, 10), 10);
+    var month = parseInt(timeStr.substring(5, 7), 10);    
+    var curDay = new Date();
+    if (year > curDay.getFullYear()){ return true};
+    if (year < curDay.getFullYear()){ return false};
+    if (month-1 > curDay.getMonth()){ return true};
+    if (month-1 <  curDay.getMonth()){ return false};
+    if (day >= curDay.getDate()){ return true};
+    return false;
```
calebsander: If you can construct the other date into a `Date` object, you can just `return date >= new Date()`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+}
+
+function showEventInfo(event)
+{
+    var data = event.data.data;
+  data = event.data.data;
```
calebsander: Unnecessary

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+  else
+  {
+    $('#start_hour').val(0);
+    $('#end_hour').val(0);
+    $('#start_minute').val(0);
+    $('#end_minute').val(0); 
+  }
+  $('#start_date').val(begin.substring(0, 10));
+  $('#end_date').val(end.substring(0, 10));
+  //$('#'+data.id).append($('#extendInfo'));
```
calebsander: Can this be removed?

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+  $('#extendInfo').show();
+}
+
+// Show all searched event
+function renderAll(eventList, tag)
+{
+    $(tag).not($('#extendInfo')).empty();
```
calebsander: Would be faster to do the `not` in CSS: `$(tag + ':not(#extendInfo)')`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+        divNode.style.backgroundColor = color_label[curEvent['organizer']['displayName']];
+        divNode.className = 'searchEvents';
+        divNode.className +=  ' rounded';
+        divNode.setAttribute("id", curEvent['id']);
+        var eventName = document.createElement('Label');
+        eventName.innerHTML = curEvent['summary'];
+        divNode.appendChild(eventName);
+        $(tag).append(divNode);
+        $("#"+curEvent.id).on('click',{data: curEvent, div:divNode}, showEventInfo);
+    });
```
calebsander, line 79: It might be more pleasant to build these elements in jQuery, e.g.
```js
var divNode = $('<div>')
  .css('background', color_label[curEvent.organizer.displayName])
  .addClass('searchEvents rounded')
  .attr('id', curEvent.id)
  .append($('<label>').text(curEvent.summary))
  .click({data: curEvent, div: divNode}, showEventInfo)
$(tag).append(divNode)
```

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+{
+    $(tag).not($('#extendInfo')).empty();
+    eventList.forEach(function(curEvent) {
+        var divNode = document.createElement('div');        
+        divNode.style.backgroundColor = color_label[curEvent['organizer']['displayName']];
+        divNode.className = 'searchEvents';
+        divNode.className +=  ' rounded';
+        divNode.setAttribute("id", curEvent['id']);
+        var eventName = document.createElement('Label');
+        eventName.innerHTML = curEvent['summary'];
```
calebsander: Do not use `innerHTML`; this allows a XSS attack

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+        divNode.appendChild(eventName);
+        $(tag).append(divNode);
+        $("#"+curEvent.id).on('click',{data: curEvent, div:divNode}, showEventInfo);
+    });
+    if ($.isEmptyObject(eventList))
```
calebsander: Could just use `if (!eventList.length)`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+{
+  var possibleEventsFuture= [];
+  var possibleEventsPast = [];
+  data['events'].forEach(function(calEvent) {
+      var name = calEvent['summary'].toLowerCase();
+      var description = '';
+      if (typeof(calEvent['description']) !== 'undefined')
+      {
+        description = calEvent['description'].toLowerCase();
+      }
```
calebsander: `var description = (calEvent.description || '').toLowerCase();`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+      var score = 0;
+      query.forEach(function(segment) {
+        if(name.includes(segment))
```
calebsander: `includes()` is a relatively modern feature of JS. I would use `indexOf()` for compatibility

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+      })
+      if (score)
+      {
+        if (isFuture(calEvent))
+          {
+            possibleEventsFuture.push({calEvent:calEvent, score:score});
+          }
+          else{
+            possibleEventsPast.push({calEvent:calEvent, score:score});
+          }
```
calebsander: `(isFuture(calEvent) ? possibleEventsFuture : possibleEventsPast).push(...)`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+		.toLowerCase()
+		.split(/[ /]+/);
+   var possibleEventsFuture= [];
+   var possibleEventsPast = [];
+   // Get all possible events...
+   $.ajax({
+      url: '/1/calendar_all_events',
+      type: 'POST',
+      data:{
+      },
```
calebsander: `data` can be omitted if it's empty

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
+var CURRENT_DATE = new Date();
+var d = new Date();
+
+var content = 'January February March April May June July August September October November December'.split(' ');
+var weekDayName = 'SUN MON TUES WED THURS FRI'.split(' ');
+var daysOfMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+var events = {}
+var curYear = 2020;
```
calebsander: Would be good not to have to update this every year. Use `CURRENT_DATE.getFullYear()`?

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  {
+    $('#start_hour').val(0);
+    $('#end_hour').val(0);
+    $('#start_minute').val(0);
+    $('#end_minute').val(0); 
+  }
+  $('#start_date').val(begin.substring(0, 10));
+  $('#end_date').val(end.substring(0, 10));
+  $(".deleteEvent").off("click");
+  $(".deleteEvent").on('click',{eventID:data.id, tag:data.organizer.displayName}, deleteEvent);
```
calebsander: Can just use `.click({eventID: data.id, tag: data.organizer.displayName}, deleteEvent)`. You can also chain the `.off()` and the `.click()` so you on't need to call `$('.deleteEvent')` twice

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  {
+    $('#extraInfo_triangle').css("left", $('#cal').width()/2 - eleCenter);
+    $('#extraInfo').css("left", $('#cal').width()/2);  
+    $('#extraInfo_triangle').css({"box-shadow": "3px -3px 3px 0px rgba(0, 0, 0, 0.4)"});
+    $('#extraInfo_triangle').css({"border-color": "white white transparent transparent"});
+  }
+  else {
+    $('#extraInfo').css("left", "0px");
+    $('#extraInfo_triangle').css("left", $('#cal').width()/2 - eleCenter);
+    $('#extraInfo_triangle').css({"box-shadow": "-3px 3px 3px 0px rgba(0, 0, 0, 0.4)"});
```
calebsander, line 96: These 2 lines seem to be the same in both cases

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  $('#extraInfo_triangle').css("top", 
+        Math.min($('#extraInfo').height() -  $('#extraInfo_triangle').height(), event.clientY+ 5));
+  $('#extraInfo_triangle').fadeIn();
+  $('#extraInfo').fadeIn();
+}
+
+// Render the events for a month and year on our calendar. 
+function renderEvents(month, year)
+{
+  $('div').filter($('.day')).each(function(){$(this).empty();});
```
calebsander: Perform the selection in CSS instead: `$('div.day')`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  for(var i = 0; i < events[month].length; i++)
```
calebsander: If you aren't using the index `i`, could do `events[month].forEach(function(curEvent) {...})`

donut/modules/calendar/static/list_script.js
```diff
@@ -0,0 +1,182 @@
...
+  'Other':'#868686',
+
+};
+function isFuture(curEvent)
+{
+
+    var timeStr = curEvent['end']['dateTime'] || curEvent['end']['date'];
+    var year = parseInt(timeStr.substring(0, 4), 10);
+    var day = parseInt(timeStr.substring(8, 10), 10);
+    var month = parseInt(timeStr.substring(5, 7), 10);
```
calebsander: Can just use `Number(timeStr.substring(5, 7))`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+
+    var begin = curEvent.start.dateTime || curEvent.start.date;
+    var end = curEvent.end.dateTime || curEvent.end.date;
+            
+    var startDay = parseInt(begin.substring(8, 10), 10);
+    var endDay = parseInt(end.substring(8, 10), 10);
+    var startMonth = parseInt(begin.substring(5, 7), 10);
+    var endMonth = parseInt(end.substring(5, 7), 10);
+    startDay = month > startMonth - 1 ? 1 : startDay;
+    endDay = month < endMonth - 1 ? daysOfMonth[month] : endDay;
```
calebsander: Might be cleaner to have a variable `var oneMonth = month + 1` and compare against that

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+// Render the events for a month and year on our calendar. 
+function renderEvents(month, year)
+{
+  $('div').filter($('.day')).each(function(){$(this).empty();});
+  for(var i = 0; i < events[month].length; i++)
+  {
+        
+    var curEvent =  events[month][i];
+    var checked = $(':checkbox').filter(
+     function( index ) {
```
calebsander: Can omit the `index` argument

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+      divNode.appendChild(eventName);
+      if (month === d.getUTCMonth())
+      {
+        document.getElementById('day ' + j).appendChild(divNode);
+        $("#"+curEvent.id+j).on('click',{data: curEvent}, showEventInfo);
+      }
+      
+      if(!checked[0].checked){
+        divNode.style.display = "none"; 
+      }
```
calebsander, line 144: I would do all of this in jQuery instead, especially attaching click handlers to the element

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+      divNode.className = 'events';
+      divNode.className += ' '+curEvent.organizer.displayName;
+      divNode.className += ' rounded-circle'; 
+      divNode.setAttribute("id", curEvent.id + j);
+      divNode.className += ' ' + curEvent.id;        
+               
+      var eventName = document.createElement('Label');
+      eventName.innerHTML = curEvent.summary;
+      divNode.appendChild(eventName);
+      if (month === d.getUTCMonth())
```
calebsander: Use `getMonth()` if `d` is in the local timezone

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+    });
+}
+
+// Cache the events locally for an entire year. 
+function process_events(data, month, year)
+{
+ data.events = data.events || [];
+
+  for(var i = 0; i < data.events.length; i++)
+  {
```
calebsander, line 178: `data.events.forEach(...)` would be cleaner

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+      beginMonth += 1;
+    }
+  }
+  renderEvents(month, year);
+}
+
+// Actually requests for the events from the server. 
+function getEvents(month, year){
+  // We don't always need to send a request. 
+  if (month === 11||month === 0 || $.isEmptyObject(events) || curYear != year)
```
calebsander: Why the special case for Jan/Dec? Shouldn't it depend on which direction you are going?

calebsander: Also this would be cleaner with an early `return` instead of wrapping the whole function in an `if`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  var $day;
+  var i = 1;
+  for (; i <= totalDays; i++) {
+    $day = $week.find('td').eq(currentDay);
+    $day.text(i);
+    if (i === currentDate) {
+      $day.addClass('today');
+    }
+
+    $day.append("</br>");
```
calebsander: `<br>`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+    var wrapper = $('<div></div>').attr('id', 'day '+i).addClass('day');
```
calebsander: No need for a closing tag; can just do `$('<div>')`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+
+// Render Calendar
+function renderCalendar(startDay, totalDays, currentDate) {
+  var currentRow = 1;
+  var currentDay = startDay;
+  var $table = $('#cal');
+  var $week = getCalendarRow();
+  var $day;
+  var i = 1;
+  for (; i <= totalDays; i++) {
```
calebsander: Why not put `var i = 1` as the `for` initialization statement?

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+    $day.text(i);
+    if (i === currentDate) {
+      $day.addClass('today');
+    }
+
+    $day.append("</br>");
+    var wrapper = $('<div></div>').attr('id', 'day '+i).addClass('day');
+    $day.append(wrapper);
+    // +1 next day until Saturday (6), then reset to Sunday (0)
+    currentDay = ++currentDay % 7;
```
calebsander: It might be cleaner to do the iteration by incrementing a `Date` object calling `getDay()` on it

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+
+    // Generate new row when day is Saturday, but only if there are
+    // additional days to render
+    if (currentDay === 0 && (i + 1 <= totalDays)) {
```
calebsander: `i < totalDays`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+      $week = getCalendarRow();
+      currentRow++;
+    }
+  }
+}
+
+// Clear generated calendar
+function clearCalendar() {
+  var $trs = $('tr').not(':eq(0)').not($('.calendar'));
```
calebsander: Should only select table rows inside `#cal`. You should convert this to a CSS query for speed improvements. I think this is equivalent to `$('#cal tbody tr:not(.calendar)')`.

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  $trs.remove();
+  $('#extraInfo').hide();
+  $('#extraInfo_triangle').hide();
+  $('.month-year').empty();
+}
+
+// Generates table row used when rendering Calendar
+function getCalendarRow() {
+  var $table = $('#cal');
+  var $tr = $('<tr/>');
```
calebsander: `/` is unnecessary; below too

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  for (var i = 0, len = 7; i < len; i++) {
```
calebsander: `for (var i = 0; i < 7; i++)`

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
+var CURRENT_DATE = new Date();
+var d = new Date();
+
+var content = 'January February March April May June July August September October November December'.split(' ');
+var weekDayName = 'SUN MON TUES WED THURS FRI'.split(' ');
+var daysOfMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
```
calebsander: Should use `Date` objects to get this information. For example `new Date(2019, 5, 0).getDate()` will give you the number of days in May 2019. Date math is notoriously hard to get right; for example, February doesn't always have 28 days.

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  if (CURRENT_DATE.getMonth() === month && CURRENT_DATE.getFullYear() === year) {
+    dateToHighlight = date;
+  }
+
+  //Getting February Days Including The Leap Year
+  if (month === 1) {
+    if ((year % 100 !== 0) && (year % 4 === 0) || (year % 400 === 0)) {
+      totalDaysOfMonth = 29;
+    }
+  }
```
calebsander: :( Use `Date` so the browser can do this for you

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+
+  // Get Start Day
+  renderCalendar(getCalendarStart(day, date), totalDaysOfMonth, dateToHighlight);
+  if (curYear != year ||$.isEmptyObject(events))
```
calebsander, line 330: This check is also being performed in `getEvents()`. Would be good to only do it in 1 place

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  });  
+
+  // Bind the little x button
+  $(document).on('click', '.close', function(){
+        $(this).parent().hide();
+        $('#extraInfo_triangle').hide();
+  });
+  $(":checkbox").each( function (){
+    this.checked = true;
+  });
```
calebsander: `$(':checkbox').prop('checked', true)`

donut/modules/calendar/static/style.css
```diff
@@ -0,0 +1,141 @@
+body{
+    padding-bottom:0%;
+    padding-top:0%;
```
calebsander: Nit: just use `0`

donut/modules/calendar/templates/add_events-iframe.html
```diff
@@ -0,0 +1,111 @@
...
+  <form method="POST" action="{{ url_for('calendar.calendar_add_events' , update=0) }}">
+         <div class="form-group">
+            <label for="name" class="required">Event Name:</label>
+         <input type="text" name="name" class="form-control" required />
+        </div>
+        <div class="form-group">
+            <label for="tag" class="required">Tag:</label>
+
+            {% for item in permissions %}
+            {% if permissions[item] and item != 'Any'%}
```
calebsander: You can move the `if` condition into the `for` statement

donut/modules/calendar/templates/add_events-iframe.html
```diff
@@ -0,0 +1,111 @@
...
+            <label for="location">Location:</label>
+            <input type="text" name="location" class="form-control"/>
+          </div>
+        <div class="form-group">
+            <label for="description" class="required">Description:</label>
+            <textarea class="form-control" name="description" rows="5" required ></textarea>
+        </div>
+      <div class="row">
+        <div class="text-center">
+          <input class="btn" type="submit" value="Submit" />
```
calebsander, line 102: Might be cleaner to use `<button>`

donut/modules/calendar/templates/calendar.html
```diff
@@ -0,0 +1,46 @@
+{% extends "layout.html" %}
```
calebsander, line 1: The indentation in this file is quite hard to read

donut/modules/calendar/templates/display-iframe.html
```diff
@@ -0,0 +1,173 @@
...
+    </tr>
+  </table>
+  {% for item in permissions %}
+    {% if item != 'Any' %}
+       <label class="checkbox-inline"><input type="checkbox" name="display_tag" value="{{ item }}">{{ item }}</label>
+    {% endif %}
+  {% endfor %}
+  <label for="jumpto" class="required" id="start_time">Jump to:
+  <input type="text" class="jumpto_month" max="12" placeholder="MM"/> </input>
+  <input type="text" class="jumpto_year" min="1970" max="2036" placeholder="YYYY"/> </input>
```
calebsander: You need `type="number"` for `min` and `max` to work

donut/modules/calendar/templates/display-iframe.html
```diff
@@ -0,0 +1,173 @@
...
+      <th>T</th>
+      <th>W</th>
+      <th>T</th>
+      <th>F</th>
+      <th>S</th>
+    </tr>
+  </table>
+  {% for item in permissions %}
+    {% if item != 'Any' %}
+       <label class="checkbox-inline"><input type="checkbox" name="display_tag" value="{{ item }}">{{ item }}</label>
```
calebsander: Generally good practice to write `<input ... />` since there's no closing tag

donut/modules/calendar/templates/display-iframe.html
```diff
@@ -0,0 +1,173 @@
...
+{% endblock %}
+{% block scripts %}
+  {{super()}}
+  <script src={{ url_for('calendar.static', filename='script.js', _external = True)}}></script>
+{% endblock %}
+{% block navbar %}{% endblock %}
+{% block content %}
+<base target="_parent" />
+  <div class="month-year text-center">
+    <h4></h4></div>
```
calebsander: Put `</div>` on a separate line?

donut/modules/calendar/static/script.js
```diff
@@ -0,0 +1,388 @@
...
+  // Generate Calendar
+  myCalendar();
+
+  // Bind visible checkboxes. 
+  $(':checkbox').change(function() {
+    toggle($(this).val(), this.checked);
+  });  
+
+  // Bind the little x button
+  $(document).on('click', '.close', function(){
```
calebsander: `$('.close').click(function() {...})`

donut/modules/calendar/templates/display-iframe.html
```diff
@@ -0,0 +1,173 @@
...
+            <textarea class="form-control" id="eventDescription_edit" name="description" rows="3" required ></textarea>
+            {% else %}
+            <label id="eventDescription" class="infoText"></label>
+            {% endif %}
+        </div>
+     {% if permissions['Any'] %}
+      <div class="row">
+        <div class="text-center">
+          <input class="btn" type="submit" value="Update" target="_parent" />
+          <input class="btn deleteEvent" value="Delete" />
```
calebsander: `<button>` would look cleaner

donut/modules/calendar/templates/share_cal-iframe.html
```diff
@@ -0,0 +1,58 @@
...
+            <select class="form-control" name="access_level">
+              <option value="reader">Read</option>
+              {% if permissions %}
+                <option value="writer">Edit</option>
+              {% endif %}
+            </select>
+          </td>
+        </div>
+      </tr>
+    </table>
```
calebsander, line 48: Is there a reason to put this in a `table`? I think Bootstrap forms are generally more readable when not in a table

sql/calendar.sql
```diff
@@ -0,0 +1,25 @@
+DROP TABLE IF EXISTS calendar_logs;
+DROP TABLE IF EXISTS calendar_events;
+CREATE TABLE calendar_logs(
```
calebsander, line 3: This table looks unused. Is it needed?

sql/calendar.sql
```diff
@@ -0,0 +1,25 @@
...
+    calendar_gmail         VARCHAR(50)  NOT NULL,
+    acl_id                 VARCHAR(50)  NOT NULL,
+    request_time           DATETIME     NOT NULL, 
+    request_permission     CHAR(4)     NOT NULL, 
+    PRIMARY KEY (log_id),
+);
+
+CREATE TABLE calendar_events(
+    event_id                 INT          NOT NULL AUTO_INCREMENT,
+    username               VARCHAR(32) , -- Who created this even
```
calebsander: Might want to clarify that this is not a Donut username

calebsander (changes requested):

sql/calendar.sql
koivyr: It is, although I was debating on if i should log it.

donut/modules/calendar/helpers.py
koivyr: yes :(

sql/calendar.sql
koivyr: It's logging everyone who requests to share a calendar with their gmail

donut/modules/calendar/static/script.js
koivyr: they should be mirrored and the code is different... I'll double check the final effect later

donut/modules/calendar/helpers.py
koivyr: Some issues with google calendar api -- there are quite a few errors that may arise (namely if there are more traffic than we expected, quota limits.) I'll change it to return on some error message from the get_events functions.

donut/modules/calendar/helpers.py
koivyr: why?

donut/modules/calendar/helpers.py
koivyr: wait why? I wrapped everything in try catch in case of google calendar api failure

donut/modules/calendar/helpers.py
calebsander: I feel like it should be the caller's responsibility to figure out what range of dates they want to query, but idk

sql/calendar.sql
calebsander: Hmm, if it's a Donut user, then I would store `user_id` instead

sql/calendar.sql
calebsander: Ah, sorry, the diff for the `helpers.py` file was hidden when I Ctrl-Fed for it.

donut/modules/calendar/helpers.py
calebsander: It's just a little dangerous to hide *all* exceptions, even if you expect some exceptions to be thrown. What if the API becomes permanently unreachable? What if there's a database error? It would be nice for the dev team to get some information about what went wrong

donut/modules/calendar/helpers.py
koivyr: u rite idk what i was thinking

donut/modules/calendar/templates/share_cal-iframe.html
koivyr: I think it looked better... I wanted it such that the inputs float to the right of the text and it was just easier for me to do it with a table

donut/modules/calendar/static/script.js
koivyr: It can't seem to find ```$('#day ' + j)``` (which is generated from this file)

calebsander: Ah, you can't have spaces in CSS `id` selectors, since the space is an operator in a CSS selector. Can you remove the space character from the id?

koivyr: ooo gotcha. Thanks!

donut/modules/calendar/static/style.css
```diff
@@ -0,0 +1,144 @@
...
+    padding-top:0;
+}
+
+.checkbox-inline input[type=checkbox]{
+    margin-left: 0;
+}
+
+#cal td{
+    height: 30%;
+    overflow: scroll;
```
calebsander: Let's change all these to `overflow: auto` so the scroll bar isn't shown if it's not needed

calebsander (approved):

PR #106, opened 2019-12-04
Restrict unauthenticated access to directory and marketplace to Caltech campus

calebsander:
### Summary
- All directory pages and marketplace pages (except the marketplace homepage) now require the user to be logged in or on Caltech internet
- Removed the ability to search the directory by hometown and stopped showing hometown on info page
- Using `login_redirect()` for marketplace rather than duplicating this functionality on a separate page

### Test Plan
- I tried visiting all the protected pages on and off the Caltech network


koivyr:
Is there a reason why you are no longer displaying hometowns?

calebsander:
I think I remember that we decided it was too stalkery, but I could be wrong

koivyr (approved):
lgtm

PR #110, opened 2020-01-28
Homepage

Thought-Weaver:
Merging the old, yet new homepage design due to additional style changes that'll be useful in redesigning the new pages.


calebsander (changes requested):
The style changes look fine, but I'd prefer it if you could revert the changes to the homepage. I think it's important to have something explaining Donut and I think loading a video every time someone visits the site is too expensive.

calebsander:
There's also a merge conflict now

calebsander (approved):

PR #111, opened 2020-01-28
Newsgroups

rlin0:
### Summary
-  Newsgroups for sending messages and viewing past messages


.travis.yml
```diff
@@ -2,6 +2,8 @@ language: python
 python:
   - "3.6"
 dist: bionic
+=======
+dist: bionic
```
rlin0: typo will fix

donut/email_utils.py
```diff
@@ -16,6 +17,20 @@ def send_email(to, msg, subject, use_prefix=True):
...
+def newsgroup_send_email(recipients, group, poster, subject, msg):
+    """
+    Sends an email to a newsgroup with recipients hidden. recipients is a list of emails.
+    group is the newsgroup name. poster is the position/person of the sender.
+    """
+
+    msg = MIMEText(msg)
+    msg['Subject'] = subject
+    msg['From'] = poster + '@donut.caltech.edu'
+    msg['To'] = group.lower().replace(' ', '_')
```
calebsander: Seems like `To` should be a valid email address

donut/email_utils.py
```diff
@@ -16,6 +17,20 @@ def send_email(to, msg, subject, use_prefix=True):
...
+    with smtplib.SMTP('localhost') as s:
+        s.sendmail('auto@donut.caltech.edu', [msg['To']] + recipients,
```
calebsander: Do we have a handler for incoming SMTP messages? If not, probably shouldn't pretend the group has an email address.

donut/email_utils.py
```diff
@@ -16,6 +17,20 @@ def send_email(to, msg, subject, use_prefix=True):
...
     msg['To'] = to
 
-    s = smtplib.SMTP('localhost')
-    s.sendmail('auto@donut.caltech.edu', [to], msg.as_string())
-    s.quit()
+    with smtplib.SMTP('localhost') as s:
+        s.sendmail('auto@donut.caltech.edu', [to], msg.as_string())
+
+
+def newsgroup_send_email(recipients, group, poster, subject, msg):
```
calebsander: This function is very similar to the `send_email()`; can you just add a `poster='auto'` default argument to `send_email()`?

donut/email_utils.py
calebsander: Also `From` should match the sender. Might want to use Python's first [`email` example](https://docs.python.org/3/library/email.examples.html) so `To` and `From` don't need to be repeated.

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+    NATURAL JOIN position_holders
+    WHERE user_id=%s 
+    AND groups.newsgroups=1
+    AND positions.send=1
+    UNION DISTINCT
+    SELECT group_name, group_id
+    FROM groups 
+    WHERE groups.newsgroups=1
+    AND groups.anyone_can_send=1
+    '''
```
calebsander: Could simplify the `UNION DISTINCT` to `... AND (send = 1 OR anyone_can_send = 1)`

sql/donut.sql
```diff
@@ -199,3 +213,12 @@ CREATE TABLE news (
...
     PRIMARY KEY (news_id)
 );
+CREATE TABLE group_applications (
+    group_app_id    INT NOT NULL AUTO_INCREMENT,
+    user_id         INT NOT NULL, 
+    group_id        INT NOT NULL, 
+    PRIMARY KEY (group_app_id),
+    FOREIGN KEY (user_id) REFERENCES members(user_id),
+    FOREIGN KEY (group_id) REFERENCES groups(group_id),
+    UNIQUE(user_id, group_id)
```
calebsander: Could just use `(user_id, group_id)` as the primary key

sql/donut.sql
```diff
@@ -15,6 +15,7 @@ DROP TABLE IF EXISTS options;
 DROP VIEW IF EXISTS members_full_name;
 DROP TABLE IF EXISTS members;
 DROP TABLE IF EXISTS buildings;
+DROP TABLE IF EXISTS newsgroup_posts;
```
calebsander, line 18: Should also drop `group_applications`

donut/templates/includes/header.html
```diff
@@ -34,7 +34,8 @@
                             <li><a href="{{ url_for('directory_search.directory_search') }}">Directory</a></li>
                             <li><a href="{{ url_for('feedback.feedback', group='donut') }}">Donut Feedback</a></li>
 			    <li><a href=" {{url_for('marketplace.marketplace') }} ">Marketplace</a></li>
-                            <li><a href="#">Newsgroups</a></li>
+                            <li><a href=" {{url_for('marketplace.marketplace') }} ">Marketplace</a></li>
```
calebsander, line 35: Marketplace appears duplicated

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+
+def unsubscribe(user_id, group_id):
+    if not user_id:
+        return None
+    query = """
+    DELETE FROM position_holders 
+    WHERE user_id = %s AND pos_id IN (
+    SELECT pos_id FROM positions NATURAL JOIN 
+    position_holders WHERE user_id = %s AND group_id = %s)
+    """
```
calebsander, line 107: Hmm, not sure unsubscribe should remove them from their position. Maybe `position_holders` should have a `receive` boolean that can be set to 0?

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+    WHERE user_id = %s AND group_id = %s
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (user_id, group_id))
+
+
+def send_email(data):
+    """Sends email to newsgroup."""
+
+    query = """
```
calebsander, line 144: Would be better to use `get_position_data()` from `groups.helpers` so we don't need to duplicate this query. You could add the ability to filter by fields like `receive`.

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+    """Gets allowed actions for user."""
+
+    if not user_id:
+        return None
+    query = """
+    SELECT p.send AS send, p.control AS control, p.receive AS receive
+    FROM groups NATURAL JOIN positions p 
+    NATURAL JOIN position_holders
+    WHERE user_id=%s AND groups.group_id=%s
+    """
```
calebsander: Should take indirect positions into account

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+def get_my_newsgroups(user_id):
+    """Gets groups user is a member of."""
+
+    query = '''
+    SELECT DISTINCT group_name, group_id 
+    FROM position_holders NATURAL JOIN positions
+    NATURAL JOIN groups
+    WHERE user_id=%s
+    AND groups.newsgroups=1
+    '''
```
calebsander: Should take indirect positions into account. Can use `get_position_data()` from `groups.helpers`.

donut/modules/newsgroups/helpers.py
calebsander: Should also take indirect positions into account

donut/modules/newsgroups/helpers.py
calebsander: Should also take indirect positions into account

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+
+def get_owners(group_id):
+    """Get users with control access to group."""
+
+    query = """
+    SELECT user_id, pos_name 
+    FROM positions NATURAL JOIN position_holders 
+    WHERE positions.control=1
+    AND positions.group_id=%s
+    """
```
calebsander, line 203: Should take indirect positions into account

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,215 @@
...
+def get_my_positions(group_id, user_id):
+    """Get positions user holds in a group."""
+
+    query = """
+    SELECT pos_name, pos_id 
+    FROM positions NATURAL JOIN position_holders
+    WHERE positions.group_id=%s
+    AND position_holders.user_id=%s
+    AND positions.send=1
+    """
```
calebsander, line 219: Should take indirect positions into account

calebsander: Also, why does this only return positions that have sending privileges?

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    user_id = auth_utils.get_user_id(flask.session['username'])
+    fields = ['group', 'subject', 'msg', 'poster']
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    data['group_name'] = groups.get_group_data(data['group'],
+                                               ['group_name'])['group_name']
+    if not data['poster']:
+        user = account.get_user_data(user_id)
+        data['poster'] = ' '.join([user['first_name'], user['last_name']])
```
calebsander: Should use `get_name_and_email()`, which takes into account preferred names

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    return flask.render_template(
+        'all_posts.html',
+        group_id=group_id,
+        group_name=group_name,
+        messages=helpers.get_past_messages(group_id, 50))
+
+
+@blueprint.route('/_delete_application')
+def delete_application(user_id, group_id):
+    #TODO
```
calebsander: Does this still need to be done? I'm okay just removing this endpoint

donut/modules/newsgroups/templates/all_posts.html
```diff
@@ -0,0 +1,19 @@
+{% extends "layout.html" %}
+{% block page %}
+{% include "navbar.html" %}
+<h1> <a href={{ url_for("newsgroups.view_group", group_id=group_id) }}>{{ group_name }}</a></h1>
+<h4> Past Messages </h4>
+    <table style="width:100%">
```
calebsander, line 6: Missing a `</table>`. Cleaning up the indentation would probably make this omission more obvious

calebsander (changes requested):

donut/email_utils.py
rlin0: To is what the email displays in the "To" header. Email should show that it's to the group instead of list of everyone's emails

donut/email_utils.py
rlin0: but newsgroup requires hiding the recipient emails

donut/modules/newsgroups/helpers.py
rlin0: It's used to return which positions the user can post as

donut/email_utils.py
calebsander: Okay I tried it and it seems like `@donut.caltech.edu` is automatically appended to the `To` field. Just make sure that the group name is not used below as a recipient.

donut/email_utils.py
calebsander: Okay, add a `recipient_group` optional parameter too?

donut/modules/newsgroups/helpers.py
calebsander: Okay, can the function be renamed so it's clearer what it's doing?

donut/modules/newsgroups/helpers.py
rlin0: but it's a lot easier to just use a separate query...

donut/modules/newsgroups/helpers.py
```diff
@@ -0,0 +1,208 @@
...
+    UPDATE position_holders 
+    SET receive = 0
+    WHERE user_id = %s AND pos_id IN (
+    SELECT p.pos_id FROM positions p LEFT JOIN position_relations pr
+    ON p.pos_id=pr.pos_id_to
+    INNER JOIN position_holders ph ON ph.pos_id=p.pos_id OR
+    pr.pos_id_from=ph.pos_id
+    WHERE ph.user_id=%s AND p.group_id=%s)
+    """
+    query += POSITIONS_QUERY + ')'
```
calebsander: What is `POSITIONS_QUERY`? I'm confused as to how this code even runs

donut/modules/newsgroups/helpers.py
calebsander: Still applies

donut/modules/newsgroups/helpers.py
calebsander: Should still take indirect positions into account

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    data = {}
+    for field in fields:
+        data[field] = flask.request.form.get(field)
+    data['group_name'] = groups.get_group_data(data['group'],
+                                               ['group_name'])['group_name']
+    if not data['poster']:
+        data['poster'] = get_name_and_email(user_id)['full_name']
+    else:
+        data['poster'] = ' '.join((data['group_name'], data['poster']))
+    if helpers.send_email(data):
```
calebsander, line 51: This should check that the user actually has sending privileges

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    user_id = auth_utils.get_user_id(flask.session['username'])
+    return flask.render_template(
+        'newsgroups.html', groups=helpers.get_my_newsgroups(user_id))
+
+
+@blueprint.route('/newsgroups/viewmsg/<post_id>')
+def view_post(post_id):
+    if 'username' not in flask.session:
+        return flask.abort(403)
+    post = helpers.get_post(post_id)
```
calebsander, line 122: Should be restricted to users in the group

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    return flask.render_template('view_post.html', post=post)
+
+
+@blueprint.route('/newsgroups/allposts/<group_id>')
+def all_posts(group_id):
+    if 'username' not in flask.session:
+        return flask.abort(403)
+    user_id = auth_utils.get_user_id(flask.session['username'])
+    group_name = groups.get_group_data(group_id, ['group_name'])['group_name']
```
calebsander, line 135: Should be restricted to users in group

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    applications = None
+    if pos_actions and pos_actions['control']:
+        applications = helpers.get_applications(group_id)
+
+    return flask.render_template(
+        'group.html',
+        group=group_info,
+        member=groups.is_user_in_group(user_id, group_id),
+        actions=pos_actions,
+        messages=helpers.get_past_messages(group_id),
```
calebsander: Should be restricted to users in group

donut/modules/newsgroups/routes.py
```diff
@@ -0,0 +1,144 @@
...
+    return flask.render_template(
+        'all_posts.html',
+        group_id=group_id,
+        group_name=group_name,
+        messages=helpers.get_past_messages(group_id, 50))
+
+
+@blueprint.route('/_delete_application')
+def delete_application(user_id, group_id):
+    helpers.remove_application(user_id, group_id)
```
calebsander, line 148: Should check that the user has admin privileges in the group

calebsander (changes requested):

donut/modules/newsgroups/helpers.py
rlin0: typo

donut/modules/newsgroups/routes.py
rlin0: Aren't groups visible to everyone as long as visible is true?

donut/modules/newsgroups/helpers.py
rlin0: Don't users with control access have to have a direct position?

donut/modules/newsgroups/routes.py
calebsander: But the messages should be hidden

donut/modules/newsgroups/helpers.py
calebsander: I don't think so? I imagine you could be house president, which gives you control access over IHC group

calebsander (approved):

PR #113, opened 2020-02-11
Lower werkzeug version

koivyr:
### Summary
- I think flask bumped up the werkzeug version since 1.0 was just released, and now secure_filename is in werkzeug.util. Not sure if we should also bump up the version (by renaming) or just cap things at 0.16.x?

### Test Plan
- make test on travis is happy


calebsander:
I just fixed the import to use `werkzeug.utils` instead, which works in both the old and new versions.

PR #114, opened 2020-02-16
Bug fix/move doc to db

koivyr:
### Summary
- Import script for historical donut content (todo -- deal with images and re-linking them)
- Store markdown in db instead of directly on the server, since there are close to 800 individual md files and this is a better way to organize them. 
- Update showdown version. 
### Test Plan
- Make test passes
- Run the local server and see that old content's there
- Yay


donut/modules/editor/helpers.py
```diff
@@ -125,27 +127,19 @@ def read_file(path):
...
+        cursor.execute(query, [])
+    res = cursor.fetchall()
+    results = {
+        key['title']: flask.url_for('uploads.display', url=key['title'])
+        for key in res
+    }
     return results
 
 
+### TODO: this functino literally has no purpose but I need to remember to get rid of it.
```
calebsander: Does this comment still apply? Please remove either the function or the comment

donut/modules/editor/routes.py
```diff
@@ -101,12 +99,13 @@ def delete_page():
     """
     filename = flask.request.args.get('filename')
     if filename != None and helpers.check_edit_page_permission():
-        helpers.remove_link(filename)
+        helpers.remove_file_from_db(filename)
 
-    return flask.redirect(flask.url_for('editor.page_list'))
+    return flask.redirect(
+        flask.url_for('editor.page_list', subdir="all_pages"))
```
calebsander: Does `subdir` actually do anything?

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+    # Headings
+    matches = re.findall(r"\=+[\+0-9a-zA-Z\(\)\'\* \-\,\?\.]+\=+", text)
+    for i in matches:
+        text = text.replace(i, i.replace("=", "#"))
+
+    # Find all links and replace them -- purposefully missing @
+    matches = re.findall(
+        r"\[[\&\~\?\%\+0-9a-zA-Z.\./\-_:\#\= ]* [\'0-9a-zA-Z.\./\-_:\# ]*\]",
+        text,
+        flags=0)
```
calebsander: `flags=0` not necessary

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+        db.commit()
+    finally:
+        db.close()
+
+
+def format_text(text):
+    # Headings
+    matches = re.findall(r"\=+[\+0-9a-zA-Z\(\)\'\* \-\,\?\.]+\=+", text)
+    for i in matches:
+        text = text.replace(i, i.replace("=", "#"))
```
calebsander, line 50: Shouldn't the `#` characters only go at the front?

calebsander: Also would be better to use `re.sub()` to avoid searching through `text` every time `replace()` is called

calebsander: Also this regular expression assumes that `[\+0-9a-zA-Z\(\)\'\* \-\,\?\.]` are the only possible characters in links. Can you use `[^=\n]` instead to match all characters until the next `=` or newline?

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+        r"\[[\&\~\?\%\+0-9a-zA-Z.\./\-_:\#\= ]* [\'0-9a-zA-Z.\./\-_:\# ]*\]",
+        text,
+        flags=0)
+    for i in matches:
+        link = i.replace("[", "").replace("]", "")
+        # External link
+        if re.match(r"https?://", link):
+            link_pieces = link.split(" ")
+            text = text.replace(i, '[' + " ".join(link_pieces[1:]).strip() +
+                                "]" + "(" + link_pieces[0].strip() + ")")
```
calebsander: If you split on `r'\s+'` instead, seems like you can skip the `strip()`s

donut/modules/editor/utils/media_wiki_import.py
calebsander: And same comments as above on `re.sub()` and not listing all possible characters (just `[^\]]` should suffice)

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+                                link_piece[0].strip().replace(" ", "_") + ')')
+        # The text is different from the link title.
+        else:
+            text = text.replace(i, '[' + link_piece[1].strip() + ']' + '(' +
+                                link_piece[0].strip().replace(" ", "_") + ')')
+    # Make bullet points
+    text = text.replace("*", "* ")
+
+    # Nested loops
+    text = text.replace("* * ", "    * ")
```
calebsander: Put this before the previous replacement to make it clearer that this is replacing `**` in the original text?

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # Bolded first --> ''' -> **
+    text = text.replace("\'\'\'", "**")
+
+    # Italics
+    text = text.replace("\'\'", "*")
```
calebsander: Backslashes unnecessary here and above

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # There's some css --- restore them
+    text = text.replace("style#", "style=")
```
calebsander: Can use a negative lookbehind in your original regex to avoid matching `style` before `=`

calebsander (changes requested):

donut/modules/editor/utils/media_wiki_import.py
koivyr: > Shouldn't the `#` characters only go at the front?

Either works I guess. I think it's clearer to have it around the heading though.

calebsander (dismissed):

koivyr:
Orz i deleted a few useless functions

donut/modules/editor/helpers.py
```diff
@@ -125,42 +126,17 @@ def read_file(path):
...
-    '''
-    links, root = get_glob()
-    results = {}
-    for filename in links:
-        link = flask.url_for('uploads.display', url=filename)
-        results[filename] = link
+    query = """SELECT title FROM webpage_files"""
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [])
+    res = cursor.fetchall()
```
calebsander: I think this line should be inside the `with` statement. Not sure whether it's safe to use the cursor after it's been closed

donut/modules/editor/utils/media_wiki_import.py
```diff
@@ -0,0 +1,127 @@
...
+                    text = format_text(text)
+                    cursor.execute(query, [title, date, text])
+        db.commit()
+    finally:
+        db.close()
+
+
+def format_text(text):
+    # Headings
+    matches = re.findall(r"\=+[\+0-9a-zA-Z\(\)\'\* \-\,\?\.]+\=+", text)
```
calebsander, line 48: Still think you could use inverted character classes to avoid listing all the possible characters, but as long as it works with the current legacy pages, I'm okay

calebsander (dismissed):

donut/modules/editor/helpers.py
koivyr: shit u rite

PR #120, opened 2020-02-22
Support ties in election rankings and add a surveys permission

calebsander:
### Summary
- Candidates for elected position questions can now be ranked with multiple candidates per ranking. Ranked pairs will only tally candidates ranked in higher rankings against candidates ranked in lower rankings.
- Add validation that no candidates are ranked multiple times, since this could be used to inflate the ranked pairs tallies
- Added a surveys permission that is needed to create or edit surveys
- Fix a SQL bug when getting permissions for users who don't hold any positions
- Add an index on `surveys.access_key` for faster lookup

### Test Plan
- Tested new changes in the web interface
- Updated tests, still at 100% coverage

### DIFF
--- a/donut/modules/voting/helpers.py
+++ b/donut/modules/voting/helpers.py
@@ -16,8 +16,6 @@ AM_OR_PM = set(['A', 'P'])
 YYYY_MM_DD = '%Y-%m-%d'
 NO = 'NO'
 ALREADY_COMPLETED = 'Already completed'
-# The number of results to list for elected positions
-WINNERS_TO_LIST = 3
 
 
 def get_groups():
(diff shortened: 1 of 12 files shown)


donut/modules/voting/helpers.py
```diff
@@ -412,5 +431,5 @@ def get_results(survey_id):
         elif question_type == question_types['Checkboxes']:
             question['results'] = Counter(chain(*responses)).most_common()
         elif question_type == question_types['Elected position']:
-            question['results'] = winners(responses)
+            question['results'] = winners(responses)[:WINNERS_TO_LIST]
```
koivyr: I think for transparency purposes it should list everyone.

koivyr (dismissed):
lgtm otherwise

donut/modules/voting/helpers.py
calebsander: Yeah I was worried that people would write in the entire Bee Movie script. But I guess that's not a problem any more since they can only write in Donut users.

PR #121, opened 2020-02-23
Improve registrar course importer

calebsander:
### Summary
- Support updating existing courses and deleting courses that no longer exist
- Add `-e` flag so courses can be imported in production environment

### Test Plan
- Ran sample imports of different versions of the 2019-20 SP export file

### DIFF
--- a/donut/modules/courses/utils/import_registrar.py
+++ b/donut/modules/courses/utils/import_registrar.py
@@ -36,135 +36,195 @@ import sys
 from donut.pymysql_connection import make_db
 from donut.modules.courses.helpers import TERMS
 
-COURSE_MATCH = r'^(?P<dept>[A-Za-z/]+) *0*(?P<number>[0-9]+[A-Z]*)$'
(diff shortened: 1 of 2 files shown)


donut/modules/courses/utils/import_registrar.py
```diff
@@ -36,135 +36,195 @@
...
+                       (course['name'], course['units_lecture'],
+                        course['units_lab'], course['units_homework']))
+        course_id = cursor.lastrowid
+        if course_id:
+            course_ids[number] = course_id
+            print('New course:', number)
+        else:
+            cursor.execute(get_course_query, course_key)
+            course_ids[number] = cursor.fetchone()['course_id']
+            print('Updated course:', number)
```
koivyr, line 127: remove the print?

calebsander: I added the prints intentionally so whoever runs the script has a sanity check that they imported the right file

Issue #115, opened 2020-02-19
Lower the case of emails when checking in forget passwords


koivyr:
using zmo@Caltech.edu gets me incorrect email/username error, when my email in the db was zmo@caltech.edu. Case shouldn't matter here.

PR #127, opened 2020-02-25
image+media importer+bugs

koivyr:
### Summary
- #115 email case shouldn't matter
- Add image parse + media parse when converting from mediawiki to markdown
- Add revcom tab + fix ESE link in header. 
- Move Account management to only show when user is not logged in -- it doesn't make sense for a user who already logged on to do 'forget password' and to do 'request an account'

### Test Plan
- make test

### DIFF
--- a/donut/modules/auth/helpers.py
+++ b/donut/modules/auth/helpers.py
@@ -60,7 +60,7 @@ def handle_forgotten_password(username, email):
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, [username])
         result = cursor.fetchone()
-
+    email = email.lower()
     if result is not None and email == result['email']:
         name = result['first_name']
         user_id = result['user_id']
--- a/donut/modules/auth/helpers.py
+++ b/donut/modules/auth/helpers.py
@@ -60,8 +60,7 @@ def handle_forgotten_password(username, email):
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, [username])
         result = cursor.fetchone()
-    email = email.lower()
-    if result is not None and email == result['email']:
+    if result is not None and email.lower() == result['email'].lower():
         name = result['first_name']
         user_id = result['user_id']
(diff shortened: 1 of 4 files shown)


donut/modules/auth/helpers.py
```diff
@@ -60,7 +60,7 @@ def handle_forgotten_password(username, email):
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, [username])
         result = cursor.fetchone()
-
+    email = email.lower()
```
calebsander: If the email in the database is not lower-cased, I think we should still send to that email. Can we just make the comparison `email.lower() == result['email'].lower()` and send the email to `result['email']` instead?

donut/templates/includes/header.html
```diff
@@ -28,12 +28,11 @@
                         <ul class="dropdown-menu" role="menu">
                             <li><a href="{{ url_for('feedback.feedback', group='arc') }}">ARC Feedback</a></li>
                             <li><a href="{{ url_for('calendar.calendar') }}">Calendar</a></li>
-                            <li><a href="#">Course Reviews</a></li>
                             <li><a href="{{ url_for('courses.planner') }}">Course Planner</a></li>
                             <li><a href="{{ url_for('courses.scheduler') }}">Course Scheduler</a></li>
                             <li><a href="{{ url_for('directory_search.directory_search') }}">Directory</a></li>
                             <li><a href="{{ url_for('feedback.feedback', group='donut') }}">Donut Feedback</a></li>
-			    <li><a href=" {{url_for('marketplace.marketplace') }} ">Marketplace</a></li>
+			                <li><a href=" {{url_for('marketplace.marketplace') }} ">Marketplace</a></li>
```
calebsander: Let's standardize the indentation to use only spaces here so this lines up better

calebsander (changes requested):
Looks good, just a few minor comments

calebsander (approved):

PR #131, opened 2020-03-03
Logging

### LINKED ISSUE #15: Logging
The logging infrastructure needs to support many use cases, such as (add more if you think of ones I missed):
- Account created
- Account creation failed (e.g. attempted to use broken/invalid link)
- User logged in
- User login failed
- Server side uncaught exceptions
- Other security-related issues (e.g. accesses to sensitive data)

While logging strings to a file is easy, it's pretty wasteful of disk space for the amount of information recorded and parsing takes nontrivial effort. Instead, if we have (one? multiple?) tables used for logging, combined with enums, it would be much easier to work with. 

Here's something like what I have in mind:

In the codebase:

``` python
class LoggedEventType(object):
  ACCOUNT_CREATED = 1
  USER_LOGIN = 2
  USER_LOGIN_FAIL = 3
  SERVER_ERROR = 4
  etc

def log_event(event_type, ...):
  # Log the event that occurred, the relevant user (if any), the IP address, etc.
```

In the database:

``` sql
CREATE TABLE event_log (
  logged_event_id INTEGER NOT NULL AUTO_INCREMENT,
  logged_event_type INTEGER NOT NULL,
  logged_event_time DATETIME NOT NULL,
  user_id INTEGER, -- Nullable
  ip_address VARCHAR(255),
  additional_notes VARCHAR(65535),
  PRIMARY KEY (logged_event_id),
  FOREIGN KEY (user_id) REFERENCES members (user_id)
);
```

Storing event information as integer enums (4 bytes) and user_ids is much more efficient than writing down the whole string like "user dkong logged in from ip address 192.168.0.1". The latter is usually done to make the logs actually readable, but if the data is stored in the database then we can create a web view of the logs that makes it readable without wasting disk space.

Edit: It may be a good idea to store user activity logs on the database but to keep server errors and such logged to a file. If something happens to the server (e.g. cannot connect to database) then we don't want to lose the logs, and filesystem is much less likely to fail than a database.

calebsander:
### Summary

Fixes #15. I decided not to use the database to store logs because it seems important to be able to get the logs if the database is corrupted, as we experienced last year.
Currently logging:
- Account routes (e.g. create new account)
- Auth routes (e.g. logging in/masquerading)
- On top of the Flask logging for every request made, the parsed query arguments, the logged-in user (if any), and the response time

### Test Plan
- Made requests to all the endpoints, verified they were logged correctly

### DIFF
--- a/donut/__init__.py
+++ b/donut/__init__.py
@@ -76,6 +76,7 @@ def init(environment_name):
 # Create database engine object.
 @app.before_request
 def before_request():
+    flask.request.start = datetime.datetime.now()
     """Logic executed before request is processed."""
     if ('DB_NAME' in app.config and 'DB_USER' in app.config
             and 'DB_PASSWORD' in app.config):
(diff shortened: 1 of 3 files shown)


koivyr (approved):
<sub><sup>Yay now we can stalk people</sup></sub>

PR #132, opened 2020-03-04
Fix #129

### LINKED ISSUE #129: Incorrect width in course scheduler
Screen Shot 2020-02-27 at 00 32 18
L174 should span but does not.

calebsander:
### Summary
Separate groups of courses by gaps in the schedule. Allow each group to have a separate number of columns so we use screen space as well as possible.

### Test Plan
Scheduled some courses:
Screen Shot 2020-03-03 at 17 45 53

### DIFF
--- a/donut/modules/courses/templates/scheduler.html
+++ b/donut/modules/courses/templates/scheduler.html
@@ -319,34 +319,74 @@
     }
     function fitIntervals() {
       dayLists.each(function(_, day) {
-        var intervals = []
+        var intervals = [], intervalBounds = []
         $(day).children('.interval').each(function(_, element) {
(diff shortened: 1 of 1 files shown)


donut/modules/courses/templates/scheduler.html
```diff
@@ -319,34 +319,74 @@ <h1>Course Scheduler</h1>
...
+            start: true,
+            interval: interval
+          }, {
+            time: interval.end,
+            start: false
           })
         })
+
+        /*
+          Use the greedy interval scheduling algorithm
```
koivyr, line 342: straight outta an interview lol

koivyr (approved):

PR #140, opened 2020-03-06
Use 1 table for all feedbacks

rlin0:
### Summary
- Combine SQL feedback tables for boc, arc, donut into 1 general table for feedback


sql/feedback.sql
```diff
@@ -0,0 +1,31 @@
...
+  UNIQUE KEY (uuid)
+);
+
+CREATE TABLE complaint_messages (
+  complaint_id INT(11) DEFAULT NULL,
+  time TIMESTAMP,
+  message TEXT,
+  poster VARCHAR(50) DEFAULT NULL,
+  message_id INT(11) NOT NULL AUTO_INCREMENT,
+  org VARCHAR(10) NOT NULL,
```
calebsander: Are all `complaint_messages` attached to a `complaint_info`? If so, seems like `org` is redundant.

sql/feedback.sql
```diff
@@ -0,0 +1,31 @@
+DROP TABLE IF EXISTS complaint_emails;
+DROP TABLE IF EXISTS complaint_messages;
+DROP TABLE IF EXISTS complaint_info;
+
+CREATE TABLE complaint_info (
+  complaint_id INT(11) NOT NULL AUTO_INCREMENT,
+  subject VARCHAR(50) DEFAULT NULL,
+  status VARCHAR(36) DEFAULT NULL,
+  uuid BINARY(16) NOT NULL,
+  org VARCHAR(10) NOT NULL,
```
calebsander: Would be more efficient to store an `INT`; could define an `IntEnum` in Python that defines the different `org` values

calebsander (changes requested):
Nice improvement! Left a few minor comments about the SQL schemas.

calebsander (approved):

PR #141, opened 2020-03-06
Pagination of directory search

### LINKED ISSUE #116: Paginate things!
Pages, directories, uploads... The most important part is probably paginating directories since otherwise, it takes quite a while for mariadb to fetch the images from DB and for the page itself to load all the images. 

Also maybe marketplace

rlin0:
### Summary
- Make search results in directory search multiple pages


donut/modules/directory_search/routes.py
```diff
@@ -58,48 +60,82 @@ def search_by_name(name_query):
...
+        total=total,
+        num_pgs=num_pgs,
+        page=page,
+        name=form['name'],
+        house_id=form['house_id'],
+        option_id=form['option_id'],
+        building_id=form['building_id'],
+        grad_year=form['grad_year'],
+        username=form['username'],
+        email=form['email'])
```
calebsander: Maybe pass `form` here instead of listing out each field?

donut/modules/directory_search/helpers.py
```diff
@@ -158,6 +158,12 @@ def execute_search(**kwargs):
...
     query += ' ORDER BY LOWER(last_name), LOWER(full_name)'
+    if 'offset' in kwargs:
+        query += ' LIMIT %s, %s'
+        substitution_arguments.append(kwargs['offset'])
+        substitution_arguments.append(kwargs['per_page'])
+    else:
+        query = 'SELECT COUNT(*) AS cnt FROM (' + query + ') sub'
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, substitution_arguments)
         return cursor.fetchall()
```
calebsander: Should use `fetchone()` with the count query. Would probably be cleaner to return the `cnt` field from this function instead of the row.

donut/modules/directory_search/routes.py
```diff
@@ -58,48 +60,82 @@ def search_by_name(name_query):
...
+        per_page=PER_PAGE)
+    if flask.request.method == 'POST':
+        total = int(
+            helpers.execute_search(
+                name=name,
+                house_id=house_id,
+                option_id=option_id,
+                building_id=building_id,
+                grad_year=grad_year,
+                username=form['username'],
```
calebsander: You can put the repeated arguments in a `dict` to pass to both calls:
```python
args = dict(name=name, house_id=house_id, ...)
helpers.execute_search(**args, ...)
```

donut/modules/directory_search/helpers.py
```diff
@@ -158,6 +158,12 @@ def execute_search(**kwargs):
         query += ' AND graduation_year = %s'
         substitution_arguments.append(kwargs['grad_year'])
     query += ' ORDER BY LOWER(last_name), LOWER(full_name)'
```
calebsander: `ORDER BY` can be omitted if just counting the results

donut/modules/directory_search/templates/search_results.html
```diff
@@ -1,10 +1,8 @@
 {% extends 'layout.html' %}
 {% block page %}
 	<div class='container theme-showcase jumbotron' style='float: none; padding-bottom: 25px'>
-		<button class='btn' onclick='history.back()'>
-			&lt; Back
-		</button>
-		<div>Your search returned {{ users|length or 'no' }} results</div>
+		<a class='btn btn-info' href='{{ url_for("directory_search.directory_search") }}'>Back</a>
```
calebsander: `history.back()` was intentional so the browser will restore all the form inputs' values

donut/modules/directory_search/templates/search_results.html
```diff
@@ -39,6 +37,25 @@
...
+			<ul id='demo' class="pagination">
+	<li class="page-item {% if page == 1 %} disabled {% endif %}">
+		<a class="page-link" {% if page != 1 %} href="{{ url_for('directory_search.search', page=page-1, total=total, num_pgs=num_pgs, name=name, house_id=house_id, option_id=option_id, building_id=building_id, grad_year=grad_year, username=username, email=email) }}" {% endif %} aria-label="Previous">
+        <span aria-hidden="true">&laquo;</span>
+        <span class="sr-only">Previous</span>
+      </a>
+    </li>
+    {% for i in range(1, num_pgs+1) %}
+    <li class="page-item {% if i == page %} active {% endif %}">
+	    <a class="page-link" href="{{ url_for('directory_search.search', page=i, total=total, num_pgs=num_pgs, name=name, house_id=house_id, option_id=option_id, building_id=building_id, grad_year=grad_year, username=username, email=email) }}">{{ i }}</a>
```
calebsander: Don't think `num_pgs` is needed

donut/modules/directory_search/templates/search_results.html
```diff
@@ -39,6 +37,25 @@
 					{% endfor %}
 				</tbody>
 			</table>
-		{% endif %}
+			<ul id='demo' class="pagination">
```
calebsander: Is `id='demo'` needed?

calebsander (changes requested):

donut/modules/directory_search/templates/search_results.html
rlin0: history.back() goes back to the last results page instead of the search form

donut/modules/directory_search/routes.py
```diff
@@ -6,6 +6,8 @@
 from donut.auth_utils import get_user_id, is_caltech_user, login_redirect
 from donut.modules.directory_search import blueprint, helpers
 
+PER_PAGE = 20
```
koivyr: would be nice to allow people to change how many per page

donut/modules/directory_search/helpers.py
rlin0: like `cursor.rowcount`?

calebsander: I mean just return `cursor.fetchone()['cnt']` instead of making the calling function pull out this field

donut/modules/directory_search/templates/search_results.html
rlin0: I think it's needed so that it doesn't have to recalculate the number of pages when it's just switching pages

calebsander: Hmm, I see. I think you should be able to calculate `num_pgs` from `total`, though. That would decrease the number of parameters that need to get passed between the client and the server.

donut/modules/directory_search/templates/search_results.html
```diff
@@ -1,44 +1,106 @@
...
-						</tr>
-					{% endfor %}
-				</tbody>
-			</table>
-		{% endif %}
-	</div>
+<div class='container theme-showcase jumbotron' style='float: none; padding-bottom: 25px'>
+	<a class='btn btn-info' href='{{ url_for("directory_search.directory_search") }}'>Back</a>
+	<div>Your search returned {{ total }} results</div>
+	{% if users %}
```
calebsander, line 6: Can you fix the indentation? It's kinda hard to read what HTML and Jinja2 directives are nested inside what others. If you want to use spaces instead of tabs, that's fine.

donut/modules/directory_search/templates/search_results.html
```diff
@@ -1,44 +1,106 @@
...
+	{% if num_pgs > 15 %}
+	{% if page - 6 > 1 %}
+	<li class="page-item">
+		<a class="page-link" href="{{ url_for('directory_search.search', page=1, per_page=per_page, total=total, **query_info) }}">1</a>
+	</li>
+	<li class="page-item">
+		<a class="page-link">...</a>
+	</li>
+	{% endif %}
+	{% for i in range([1, page - 6] | max , page + 6) %}
```
calebsander: Shouldn't this be `min(page + 6, num_pgs) + 1`?

donut/modules/directory_search/templates/search_results.html
```diff
@@ -1,44 +1,106 @@
...
+</ul>
+{% endif %}
+</div>
+{% endblock %}
+{% block scripts %}
+{{ super() }}
+<script>
+	$(document).ready(function() {
+		$('#per_page').change(function() {
+			var p = $( "#per_page option:selected" ).text();
```
calebsander: What's wrong with `$(this).val()`?

calebsander (changes requested):

donut/modules/directory_search/templates/search_results.html
rlin0: It turns into the variable `i`.

calebsander: Oh, because you need to use `value={{ i }}` in the Jinja template

rlin0:
The git commit history is lowkey messed up my bad idk what happened

calebsander (approved):

PR #142, opened 2020-03-06
Positions fixed

calebsander:
### Summary
- Fixed position administration actions:
  - Administration forms actually have submit buttons now lol
  - Administration routes actually check that the user has control permissions
  - Add a form for ending the term of a current position holder
  - Removed the "delete position" form since I think we want to keep historical information about positions
  - Can now specify send/control/receive privileges when creating new positions
  - Show all positions in groups that the user controls instead of having to select the group to control first
  - Give more useful error messages
- Fix XSS vulnerabilities in positions page due to building HTML strings with unescaped positions data (yikes)
- UI improvements to campus positions page: show the group that positions belong to in the dropdown, order holders by group name and position name, use a Bootstrap table, exclude house memberships from query so page loads faster
- Made a `current_position_holders` view that checks position start and end dates, and also includes indirect positions. This should be the go-to for position queries. I have switched out all the `position_holders` queries with this view so the logic is consistent.
- Renamed `position_holders.receive` to `position_holders.subscribed` so that `NATURAL JOIN`s with `positions` behave as expected
- Add a `house_positions` view to make selecting for/excluding house membership positions easier. Removed the `group_house_membership` view since it can easily be emulated with `house_positions NATURAL JOIN current_position_holders`.
- Renamed `group_houses` to `house_groups` because I could never remember the table name. Sorry, Rachel :(
- Fix newsgroups unsubscribe route, though I don't think this is actually accessible?

### Test Plan
- Tests passing
- Thoroughly tested all the position administration actions and the SQL queries that changed

### DIFF
--- a/sql/donut.sql
+++ b/sql/donut.sql
@@ -221,15 +221,14 @@ CREATE VIEW current_position_holders AS (
     SELECT * FROM current_direct_position_holders
     UNION ALL
     SELECT
-        holders.hold_id,
-        relations.pos_id_to AS pos_id,
-        holders.user_id,
-        holders.start_date,
-        holders.end_date,
-        holders.subscribed
-    FROM current_direct_position_holders holders
-        JOIN position_relations relations
-        ON holders.pos_id = relations.pos_id_from
+        current_direct_position_holders.hold_id,
+        position_relations.pos_id_to AS pos_id,
+        current_direct_position_holders.user_id,
+        current_direct_position_holders.start_date,
+        current_direct_position_holders.end_date,
+        current_direct_position_holders.subscribed
+    FROM current_direct_position_holders
+        JOIN position_relations ON pos_id = pos_id_from
 );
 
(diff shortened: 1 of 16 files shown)


koivyr:
Damn that’s a lot of stuff
Thanks for all this lol

sql/donut.sql
```diff
@@ -164,7 +166,8 @@ CREATE TABLE positions (
     PRIMARY KEY (pos_id),
     FOREIGN KEY (group_id)
         REFERENCES groups(group_id)
-        ON DELETE CASCADE
+        ON DELETE CASCADE,
+    UNIQUE (group_id, pos_name)
```
koivyr, line 170: while we are at this, can we make position_permissions also have unique(pos_id, permission_id)

koivyr (dismissed):

koivyr (approved):

PR #145, opened 2020-03-10
Fix email x2

koivyr:
### Summary
- Fix email for the second time...

### Test Plan
-

### DIFF
--- a/donut/email_utils.py
+++ b/donut/email_utils.py
@@ -1,6 +1,5 @@
 import smtplib
 from email.mime.text import MIMEText
-from email.mime.multipart import MIMEMultipart
 
 
 def send_email(to, text, subject, use_prefix=True, group=None):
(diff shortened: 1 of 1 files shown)


donut/email_utils.py
```diff
@@ -10,7 +10,7 @@ def send_email(to, text, subject, use_prefix=True, group=None):
     is not none, the email is sent to a newsgroup and the to emails are hidden.
     """
 
-    msg = MIMEText(msg)
+    msg = MIMEText(text, "html")
```
calebsander: We need to escape this in case they use `<`, `>`, or `&` in emails. How about running newsgroup emails through a Markdown converter? This would allow people to send formatted emails without writing raw HTML, and would automatically take care of escapes.

calebsander: Also, can you remove the `MIMEMultipart` import as part of this PR?

rlin0: also remove the MIMEMultipart import lol

calebsander (changes requested):

rlin0 (dismissed):
thanks ziyan

donut/email_utils.py
Wingfril: > We need to escape this in case they use `<`, `>`, or `&` in emails. How about running newsgroup emails through a Markdown converter? This would allow people to send formatted emails without writing raw HTML, and would automatically take care of escapes.

lol is it time to use showdown.js again

koivyr: I completely forgot about xss attacks lol. 
I think I'll just stick with text/plain until later.

calebsander (approved):

PR #146, opened 2020-03-11
Fix a lot of css

koivyr:
### Summary
- Font size is changed back to 16 because we aren't going blind just yet
- Added titles to things because that's mildly useful. 
- For newgroups where 'anyone_can_send' is true, make it so that it really is anyone can send. 
- (i think newsgroup is very NOT mobile-friendly but i'll get to that some day)
- Changed the donut logo because we aren't in the early 2000's anymore. 
- make it so that when the donut-legacy media-wiki exports directly link back to the donut.caltech.edu page, when we import it, it leads to the new page instead.
- Added padding because it's usually not standard for non-mobile webpages to go all the way to the edge.  

- There are also multiple orange colors used (#f5903b vs orange vs #FF8C00)???? I didn't change anything with that but it lowkey bugs me when there are two different ones on the same site. 
### Test Plan
-


donut/modules/newsgroups/routes.py
```diff
@@ -112,7 +114,9 @@ def mygroups():
         return flask.abort(403)
     user_id = auth_utils.get_user_id(flask.session['username'])
     return flask.render_template(
-        'newsgroups.html', groups=helpers.get_my_newsgroups(user_id))
+        'newsgroups.html',
+        groups=helpers.get_my_newsgroups(user_id),
+        page_title="My Newgroups")
```
koivyr: typo will fix

donut/modules/newsgroups/templates/navbar.html
```diff
@@ -1,11 +1,10 @@
 <div class="container theme-showcase" role="main">
-<div class="btn-group" role="group">
-<div class="container theme-showcase" role="main">
+<div role="group">
 <div id="newsgroup-nav-bar">
+<h1>Newgroup Services</h1>
```
rlin0: typo

koivyr: typing is hard

donut/modules/newsgroups/helpers.py
```diff
@@ -65,14 +65,22 @@ def get_user_actions(user_id, group_id):
...
         res = cursor.fetchall()
     actions = {'send': False, 'control': False, 'receive': False}
-    if not res:
+    anyone_can_send_query = """SELECT anyone_can_send FROM groups
+        WHERE group_id=%s"""
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(anyone_can_send_query, (group_id))
+        anyone_can_send = cursor.fetchone()
+
+    if not res and not anyone_can_send:
```
calebsander, line 74: Probably don't need this special case, but doesn't really matter

calebsander (dismissed):

donut/modules/newsgroups/helpers.py
koivyr: It bugged me when the dropdown had a certain newsgroup and then when you tried to send something, it gives a permission error @.@

calebsander: No, I just meant that the `if not res and not anyone_can_send: return actions` is not necessary

PR #148, opened 2020-03-12
fix error when no groups

rlin0:
### Summary
-  When there's no groups to post to, hides the post message
- minor style changes

### DIFF
--- a/donut/modules/newsgroups/templates/post.html
+++ b/donut/modules/newsgroups/templates/post.html
@@ -3,17 +3,16 @@
 {% include "navbar.html" %}
 <div class="container theme-showcase" role="main">
 <div class="newsgroup-frame">
+{% if not groups %}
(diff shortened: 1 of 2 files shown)


donut/modules/newsgroups/templates/post.html
```diff
@@ -41,36 +37,41 @@ <h3 id='warning'></h3>
...
+		})
+	}
 
-$(document).ready(function() {
-	groupSelect = $('#group').change(loadGroup)
-	posList = $('#poster')
-	loadGroup()
-})
+	$(document).ready(function() {
+		if( $('#group').has('option').length > 0 ) {
```
calebsander: Can we do this in Python land? `{% if groups %} ... {% else %} ... {% endif %}`

calebsander (changes requested):

calebsander (approved):

PR #151, opened 2020-03-21
newsgroup fixes

### LINKED ISSUE #150: Directory names dropdown will not close on unfocus
image

When I search for a name like "alex", but don't want to select a specific name, the name suggestions dropdown will not close if i click away, which blocks the menu below

on windows 10, chrome

rlin0:
### Summary
- replaced page_title with highlighting the button on the page selected for newsgroups
- fixed the to email to use the group name not id
- use poster name as from in emails (e.g. "ug Admin")
- fixed directory search to allow closing the dropdown (fixes #150)

### DIFF
--- a/donut/email_utils.py
+++ b/donut/email_utils.py
@@ -2,7 +2,7 @@ import smtplib
 from email.mime.text import MIMEText
 
 
-def send_email(poster, to, text, subject, use_prefix=True, group=None):
+def send_email(to, text, subject, use_prefix=True, group=None, poster=''):
     """
     Sends an email to a user. Expects 'to' to be a comma separated string of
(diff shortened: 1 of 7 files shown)


donut/email_utils.py
```diff
@@ -2,7 +2,7 @@
 from email.mime.text import MIMEText
 
 
-def send_email(to, text, subject, use_prefix=True, group=None):
+def send_email(poster, to, text, subject, use_prefix=True, group=None):
```
calebsander: There are other places where we call `send_email()` in the code, e.g. for creating an account. Can you make this an optional parameter so that those pieces of code still work?

calebsander (changes requested):

calebsander (approved):

PR #153, opened 2020-03-26
Email exceptions & errors to admins

koivyr:
### Summary
- Notify donut admins when exceptions occur. 
- Fixed an error where I forgot that redirect_login is a part of auth_utils :(

### DIFF
--- a/donut/__init__.py
+++ b/donut/__init__.py
@@ -6,6 +6,7 @@ import pdb
 import traceback
 import http
 import datetime
+import logging
 
 try:
     from donut import config
@@ -13,6 +14,8 @@ except ImportError:
     from donut import default_config as config
 from donut import constants
 from donut.modules import account, auth, marketplace, calendar, core, courses, directory_search, editor, feedback, groups, newsgroups, rooms, uploads, voting
+from donut.donut_SMTP_handler import DonutSMTPHandler
+from donut.email_utils import DOMAIN
 
 app = flask.Flask(__name__)
 Bootstrap(app)  # enable Bootstrap in Flask
@@ -73,6 +76,18 @@ def init(environment_name):
     app.jinja_env.globals.update(
         current_year=lambda: datetime.datetime.now().year)
 
+    mail_handler = DonutSMTPHandler(
+        mailhost='localhost',
+        fromaddr='server-error@' + DOMAIN,
+        toaddrs=[],
+        subject='Donut Server Error',
+        db_instance=make_db())
+    mail_handler.setLevel(logging.ERROR)
+    mail_handler.setFormatter(
+        logging.Formatter(
+            '[%(asctime)s] %(levelname)s in %(module)s: %(message)s'))
+    app.logger.addHandler(mail_handler)
+
 
 # Create database engine object.
 @app.before_request
@@ -81,15 +96,19 @@ def before_request():
     """Logic executed before request is processed."""
     if ('DB_NAME' in app.config and 'DB_USER' in app.config
             and 'DB_PASSWORD' in app.config):
-        flask.g.pymysql_db = pymysql.connect(
-            host='localhost',
-            database=app.config['DB_NAME'],
-            user=app.config['DB_USER'],
-            password=app.config['DB_PASSWORD'],
-            db='db',
-            autocommit=True,
-            charset='utf8mb4',
-            cursorclass=pymysql.cursors.DictCursor)
+        flask.g.pymysql_db = make_db()
+
+
+def make_db():
+    return pymysql.connect(
+        host='localhost',
+        database=app.config['DB_NAME'],
+        user=app.config['DB_USER'],
+        password=app.config['DB_PASSWORD'],
+        db='db',
+        autocommit=True,
+        charset='utf8mb4',
+        cursorclass=pymysql.cursors.DictCursor)
 
 
 @app.after_request
--- a/donut/modules/groups/routes.py
+++ b/donut/modules/groups/routes.py
@@ -116,8 +116,8 @@ def create_pos_holder(pos_id):
     try:
         helpers.create_position_holder(pos_id, user_id, start_date, end_date)
     except Exception as e:
-        flask.current_app.logger.error('Failed to add position holder:')
-        flask.current_app.logger.exception(e)
+        flask.current_app.logger.exception('Failed to add position holder:\n' +
+                                           e)
         return jsonify({
             'success': False,
             'message': 'Unable to add position holder'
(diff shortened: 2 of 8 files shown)


donut/DonutSMTPHandler.py
```diff
@@ -0,0 +1,28 @@
...
+        self.toaddrs = self.getAdmins()
+        super().emit(record)
+
+    def getAdmins(self):
+        ''' Returns current members in Devteam '''
+        db = make_db(
+            self.environment if hasattr(self, 'environment') else "dev")
+        query = '''SELECT DISTINCT email FROM
+            members NATURAL JOIN current_position_holders NATURAL JOIN positions NATURAL JOIN groups 
+            WHERE group_name = "Devteam"
```
calebsander: Can we make this a constant?

donut/__init__.py
```diff
@@ -73,6 +76,18 @@ def init(environment_name):
     app.jinja_env.globals.update(
         current_year=lambda: datetime.datetime.now().year)
 
+    mail_handler = DonutSMTPHandler(
+        mailhost='localhost',
+        fromaddr='server-error@beta.donut.caltech.edu',
```
calebsander: Declare this as a constant somewhere so we can change it once we're out of beta?

donut/__init__.py
```diff
@@ -73,6 +76,18 @@ def init(environment_name):
...
+        toaddrs=[],
+        subject='Application Error')
```
calebsander: Can we clarify that it's a Donut error?

donut/__init__.py
```diff
@@ -73,6 +76,18 @@ def init(environment_name):
...
+    mail_handler.setLevel(logging.ERROR)
+    mail_handler.setFormatter(
+        logging.Formatter(
+            '[%(asctime)s] %(levelname)s in %(module)s: %(message)s'))
+    mail_handler.setEnv(environment_name)
```
calebsander: Why not just pass this into the constructor?

donut/DonutSMTPHandler.py
```diff
@@ -0,0 +1,28 @@
...
+        Overrides SMTPHandler's emit such that we dynamically
+        get current donut dev team members
+        '''
+        self.toaddrs = self.getAdmins()
+        super().emit(record)
+
+    def getAdmins(self):
+        ''' Returns current members in Devteam '''
+        db = make_db(
+            self.environment if hasattr(self, 'environment') else "dev")
```
calebsander: Rather than creating a new DB each time, how about abstracting the code in `before_request()` into a separate `get_db()` function and just passing an instance of that into the `DonutSMTPHandler` constructor?

donut/modules/groups/routes.py
```diff
@@ -116,8 +116,8 @@ def create_pos_holder(pos_id):
     try:
         helpers.create_position_holder(pos_id, user_id, start_date, end_date)
     except Exception as e:
-        flask.current_app.logger.error('Failed to add position holder:')
-        flask.current_app.logger.exception(e)
+        flask.current_app.logger.exception('Failed to add position holder:\n' +
+                                           e)
```
calebsander: Is there a reason for changing this? `logger.exception()` is useful because it actually logs the stack trace, whereas converting the error to a string only preserves the error message. I also don't think it's valid to concatenate a string with an exception object?

donut/__init__.py
```diff
@@ -6,13 +6,16 @@
 import traceback
 import http
 import datetime
+import logging
+from logging.handlers import SMTPHandler
```
calebsander: These imports seem unused

donut/__init__.py
```diff
@@ -6,13 +6,16 @@
...
 
 try:
     from donut import config
 except ImportError:
     from donut import default_config as config
 from donut import constants
 from donut.modules import account, auth, marketplace, calendar, core, courses, directory_search, editor, feedback, groups, newsgroups, rooms, uploads, voting
+from donut.DonutSMTPHandler import DonutSMTPHandler
```
calebsander: Nit: Python module names are usually `snake_case`

donut/__init__.py
calebsander: Also, if this sends from `beta.`, `email_utils` should probably do the same. Maybe declare a `DOMAIN` constant in `email_utils`?

calebsander (changes requested):

donut/DonutSMTPHandler.py
koivyr: What do you mean?
Jk you meant devteam

donut/modules/groups/routes.py
koivyr: So I combined them because otherwise you get two emails sent to you lol. If y’all dont mind that, then I can revert it

koivyr: Wait also [Exception info is added to the logging message](https://docs.python.org/3/library/logging.html#logging.Logger.exception) -- the stack trace also appended to the end of the message

calebsander: Okay yeah putting them in the same email is very reasonable. I get this error when trying to concatenate a string with an exception, though:
```
Python 3.7.3 (default, Dec 13 2019, 19:58:14)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> e = Exception('abc')
>>> 'some text:\n' + e
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "Exception") to str
```

koivyr: orz i'll format them

calebsander: But I think formatting also only keeps the error message?
```
>>> e = Exception('abc')
>>> f'some text:\n{e}'
'some text:\nabc'
```

Wingfril: Well, format would be something like what I had in calendars, 
` f'Error while sharing {calendar_names} with {email} by {username}: \n{e}'`. The exception itself is formatted within the message, and then the `flask.current_app.logger.exception` appends the stack trace at the end like so
```
[2020-03-25 15:37:24,951] ERROR in helpers: Calendar error: HiThere
Traceback (most recent call last):
  File "/home/ziyan/donut/donut/modules/calendar/helpers.py", line 183, in sync_data
    raise NameError('HiThere')
NameError: HiThere```

Comment (koivyr) [reply donut/modules/groups/routes.py]: We only want the error message for this portion right? For example, in calendars, I formatted like 
``` 
flask.current_app.logger.exception(
            f'Error while sharing {calendar_names} with {email} by {username}: \n{e}')
``` 
which will email us
```
[2020-03-25 15:59:47,304] ERROR in helpers: Error while sharing ASCIT,Other with moziyan414@gmail.com by wingfril:
HiThere
Traceback (most recent call last):
  File "/home/ziyan/donut/donut/modules/calendar/helpers.py", line 375, in share_calendar
    raise NameError('HiThere')
NameError: HiThere
```

The error message itself is formatted and passed in as an argument to `flask.current_app.logger.exception`, and the stack track is appended to the end of that message.

calebsander (approved):

PR #157, opened 2020-04-02
HTML emails

### LINKED ISSUE #123: Support Html markdowns in emails
ASCIT request

rlin0:
### Summary
- support html for emails
- display plain text on website

### Test Plan
- sent email to undergrads whoops

### DIFF
--- a/donut/email_utils.py
+++ b/donut/email_utils.py
@@ -1,16 +1,29 @@
 import smtplib
 from email.mime.text import MIMEText
+from email.mime.multipart import MIMEMultipart
 
 DOMAIN = "beta.donut.caltech.edu"
 
 
-def send_email(to, text, subject, use_prefix=True, group=None, poster=''):
+def send_email(to,
+               text,
+               subject,
+               use_prefix=True,
+               group=None,
+               poster='',
+               richtext=''):
     """
     Sends an email to a user. Expects 'to' to be a comma separated string of
     emails, and for 'msg' and 'subject' to be strings. If group
     is not none, the email is sent to a newsgroup and the to emails are hidden.
     """
-    msg = MIMEText(text)
+    if richtext:
+        msg = MIMEMultipart('alternative')
+        msg.attach(MIMEText(text, 'plain'))
+        msg.attach(MIMEText(richtext, 'html'))
+    else:
+        msg = MIMEText(text)
+
     if use_prefix and '[ASCIT Donut]' not in subject:
         subject = '[ASCIT Donut] ' + subject
 
--- a/donut/modules/newsgroups/routes.py
+++ b/donut/modules/newsgroups/routes.py
@@ -149,10 +149,11 @@ def all_posts(group_id):
         messages=helpers.get_past_messages(group_id, 50))
 
 
-@blueprint.route('/_delete_application')
+@blueprint.route('/_delete_application/<user_id>/<group_id>')
 def delete_application(user_id, group_id):
-    actions = helpers.get_user_actions(user_id, group_id)
-    if not actions['control']:
+    username = flask.session.get('username')
+    if username is None or not helpers.get_user_actions(
+            auth_utils.get_user_id(username), group_id)['control']:
         return flask.abort(403)
     helpers.remove_application(user_id, group_id)
     return flask.redirect(
--- a/donut/modules/newsgroups/routes.py
+++ b/donut/modules/newsgroups/routes.py
@@ -4,8 +4,12 @@ from donut.modules.groups import helpers as groups
 from donut.modules.account import helpers as account
 from donut.modules.core.helpers import get_name_and_email
 from donut import auth_utils
+import html2text
 
 from . import blueprint, helpers
+
+# Set up converter for html to plain text
+html2plain = html2text.HTML2Text()
 
 
 @blueprint.route('/newsgroups')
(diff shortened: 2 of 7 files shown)


donut/modules/newsgroups/routes.py
```diff
@@ -4,6 +4,7 @@
 from donut.modules.account import helpers as account
 from donut.modules.core.helpers import get_name_and_email
 from donut import auth_utils
+import html2text
```
calebsander, line 7: This needs to be added to `requirements.txt`

donut/email_utils.py
```diff
@@ -1,14 +1,29 @@
...
     emails, and for 'msg' and 'subject' to be strings. If group
     is not none, the email is sent to a newsgroup and the to emails are hidden.
     """
-    msg = MIMEText(text)
+    if richtext:
+        msg = MIMEMultipart('alternative')
+        part1 = MIMEText(text, 'plain')
+        part2 = MIMEText(richtext, 'html')
+        msg.attach(part1)
+        msg.attach(part2)
```
calebsander: `part1` and `part2` aren't great variable names; can you just put the `MIMEText(...)` directly into the `attach()`?

donut/email_utils.py
```diff
@@ -1,14 +1,29 @@
 import smtplib
```
calebsander, line 1: pls merge/rebase in Ziyan's changes to this file

donut/modules/newsgroups/routes.py
```diff
@@ -49,6 +50,11 @@ def post_message():
         data['poster'] = get_name_and_email(user_id)['full_name']
     else:
         data['poster'] = ' '.join((data['group_name'], data['poster']))
+    # Convert to plain text
+    h = html2text.HTML2Text()
```
calebsander: Make this object global to avoid rebuilding it every time? And `h` is not a great variable name.

donut/modules/newsgroups/routes.py
```diff
@@ -49,6 +50,11 @@ def post_message():
...
+    h.escape_snob = True
+    h.ignore_emphasis = True
```
calebsander: Is there a reason for ignoring emphasis? Seems like it could be useful.

donut/modules/newsgroups/templates/post.html
```diff
@@ -1,5 +1,17 @@
 {% extends "layout.html" %}
 {% block page %}
+  <script src='https://cdn.tiny.cloud/1/n8e5bron77mbkmlvy2i3tojaajwdsawb0b5kvet27prot9fq/tinymce/5/tinymce.min.js' referrerpolicy="origin">
```
calebsander: Hmm could we use Showdown like `editor` does to avoid this additional dependency? Then we could also store the Markdown in the database to provide styled messages on the newsgroup page. We would need to make sure they don't put raw HTML tags in Markdown, which I think just requires replacing `&` with `&amp;`, `<` with `&lt;`, and `>` with `&gt;`. (This would also let us get rid of the `html2text` dependency.)

donut/modules/newsgroups/templates/post.html
```diff
@@ -1,5 +1,17 @@
 {% extends "layout.html" %}
 {% block page %}
```
calebsander: `{% block scripts %}`

calebsander (changes requested):

donut/modules/newsgroups/templates/post.html
rlin0: Are there other downsides besides additional dependencies? Because tinymce supports colors and nicer UI than editor, and can provide styled messages on the newsgroup page as long as we insert the html message in the db.

calebsander: Okay, that's fair. Forgot about styling

calebsander: Don't think we want to put HTML in the DB because of XSS concerns

donut/modules/newsgroups/templates/post.html
rlin0: wdym?

donut/modules/newsgroups/routes.py
rlin0: Then if html email isn't supported the plain text email shows asterisks in the text. Is that preferred?

donut/modules/newsgroups/templates/post.html
calebsander: Put JS in the `{% block scripts %}` block pls

donut/modules/newsgroups/routes.py
calebsander: Well yeah, it's Markdown. Bold text, code snippets, etc. will all be in Markdown format. I think that's fine; the asterisks aren't intrusive and they indicate that there was emphasis in the original message.

donut/modules/newsgroups/templates/view_post.html
```diff
@@ -13,7 +13,18 @@ <h3>
...
 </div>
 </div>
 {% endblock %}
+{% block scripts %}
+	{{ super() }}
+	<script src="https://cdn.rawgit.com/showdownjs/showdown/1.8.7/dist/showdown.min.js"></script>
+	<script>
+		var target = $("#msg");
+		var post = {{ post.message | tojson }};
+      		converter = new showdown.Converter({strikethrough: true, tables: true});
```
koivyr: Usually better practice declare the converter, ie`var converter`
Also, indent's off

calebsander (dismissed):

calebsander (approved):

PR #165, opened 2020-04-17
Add page edit permissions for BoC

calebsander:
Also make groups utils files executable (so you don't need to put `python` in front of them)


koivyr (approved):
damn i was literally also gonna make a pr for this lol

PR #168, opened 2020-04-29
fix to feedback sending email

rlin0:
### Summary
- Fix bug with data['plain']
- Moved summary button to top (was in a weird place)

### Test Plan
- Tested with arc feedback

### DIFF
--- a/donut/modules/feedback/helpers.py
+++ b/donut/modules/feedback/helpers.py
@@ -54,6 +54,8 @@ def send_to_group(group, data):
     data['group'] = group_id
     data['group_name'] = group
     data['poster'] = "{} Feedback".format(group)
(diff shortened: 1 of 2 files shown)


donut/modules/feedback/helpers.py
```diff
@@ -54,6 +54,7 @@ def send_to_group(group, data):
     data['group'] = group_id
     data['group_name'] = group
     data['poster'] = "{} Feedback".format(group)
+    data['plain'] = data['msg']
```
calebsander, line 57: Seems like `data['msg']` should be set to `''` or `None` since it's currently being used as the HTML content of the message

calebsander (approved):

PR #169, opened 2020-05-07
ug-* admins shouldn't receive emails for the group

calebsander:
Minor fix, just prevents ug group admins from receiving emails sent to the group if they aren't also members. They should still be able to see the messages on Donut. I already made the changes to the prod DB.


koivyr:
when is this true...?

calebsander:
Well all of Devteam, BoD, ARC, RevComm, and the Tech are admins of `ug` and `ug-*`. This isn't a problem with `ug`, since everyone gets those emails anyways, but it is for the individual years. For example, if anyone sends an email to the seniors, we all receive it.

koivyr (approved):

PR #177, opened 2020-05-18
Show feedback emails

koivyr:
### Summary
- Should show all email addresses that the poster has when forwarding to members. 

### Test Plan
-


rlin0 (changes requested):
Would be nice to include the link too

calebsander:
Yeah I agree with Rachel. It's probably more helpful to just include the link to the feedback, since that page has the email addresses anyways.

rlin0 (dismissed):
Personally think including just the link is sufficient

calebsander (dismissed):
Seems fine, but I agree with Rachel that the link alone is probably fine

PR #191, opened 2020-07-08
Don't remove ug-* groups

calebsander:
### Summary
- Keep `ug-*` graduation year groups and remove their members instead of removing the groups themselves. This fixes issues where groups can't be removed because newsgroup messages have been sent to them.
- All graduation years in Donut now have `ug-*` groups, not just the current 4 years
- `ug-*` admin positions are hidden in Directory since they are implied by the `ug` admin position

### Test Plan
- Ran the script in dev and prod and successfully updated the `ug-*` groups

### DIFF
--- a/donut/modules/directory_search/helpers.py
+++ b/donut/modules/directory_search/helpers.py
@@ -69,6 +69,7 @@ def get_user(user_id):
             SELECT group_name, pos_name
             FROM current_position_holders NATURAL JOIN positions NATURAL JOIN groups
             WHERE pos_id NOT IN (SELECT pos_id FROM house_positions)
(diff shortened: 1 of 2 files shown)


rlin0:
Can't `ON DELETE CASCADE` solve the issue of groups not being able to be removed?

calebsander:
Yeah but that would delete all messages sent to the newsgroups, which I don't think we want

rlin0:
Isn't it ok to delete messages to old newsgroups? Would we want to keep messages from like 10 years ago? (assuming donut lasts that long lmao)

calebsander:
I'm not sure what our data retention policies should be, but I think we should keep things around unless we are running out of storage space (which seems unlikely to happen soon). This script needs to run every year and I think we definitely want messages to last longer than that.

PR #196, opened 2020-07-20
Migrate from Travis CI to GitHub Actions

### LINKED ISSUE #195: Consider using GitHub Actions instead of Travis
https://docs.github.com/en/actions/building-and-testing-code-with-continuous-integration/setting-up-continuous-integration-using-github-actions

calebsander:
### Summary
Fixes #195.
This doesn't change how the CI process works, but it does keep it all within GitHub, which seems convenient. It was pretty easy to set up and it seems very extensible (there are lots of open-source recipes for steps to add to the build/test workflow). We could also add other workflows later, for example automatically running `make lint` on each push. The runners also seem to boot up faster than Travis and the workflow runs about 30% faster.
I can't think of any downsides to this, except that the workflow is a little more complicated because it's running a MariaDB instance inside a separate Docker container. It only took me 18 tries to get it working 😢 .


koivyr (approved):
such hard workers y'all

PR #198, opened 2020-07-26
Embed airport carpools spreadsheet 

rlin0:
### Summary
- Add carpool spreadsheet page so that don't have to circulate it through email every break 
- ASCIT members and admins can update the link to the spreadsheet and hide the spreadsheet when it's not break season
- Users can edit the spreadsheet directly on Donut or open it in Google Sheets 

### Test Plan
- Tried on local host

### DIFF
--- a/donut/modules/flights/routes.py
+++ b/donut/modules/flights/routes.py
@@ -1,0 +1,35 @@
+import flask
+from donut import auth_utils
+
+from . import blueprint, helpers
+
+
+@blueprint.route('/flights')
+def flights():
+    if not auth_utils.check_login():
+        return flask.abort(403)
+    settings = helpers.get_settings()
+    return flask.render_template(
+        'flights.html',
+        visible=settings['visible'],
+        link=settings['link'],
+        admin=helpers.is_admin())
+
+
+@blueprint.route('/flights/update', methods=['POST'])
+def update():
+    if not helpers.is_admin():
+        return flask.abort(403)
+    link = flask.request.form.get('link')
+    visible = 'visible' in flask.request.form
+    helpers.update(link, visible)
+    return flask.redirect(flask.url_for('flights.settings'))
+
+
+@blueprint.route('/flights/settings')
+def settings():
+    if not helpers.is_admin():
+        return flask.abort(403)
+    settings = helpers.get_settings()
+    return flask.render_template(
+        'settings.html', visible=settings['visible'], link=settings['link'])
--- a/donut/modules/flights/templates/flights.html
+++ b/donut/modules/flights/templates/flights.html
@@ -1,0 +1,19 @@
+{% extends "layout.html" %}
+{% block page %}
+
+<div class="container theme-showcase" role="main">
+	<div class="jumbotron">
+		<h1>Airport Carpools</h1>
+		{% if admin %}                        
+			<a class="btn btn-primary" href="{{ url_for('flights.settings') }}" role="button">Settings</a>
+		{% endif %}
+		{% if visible %}
+			<a class="btn btn-primary" href="{{ link }}" target="_blank">Go to Spreadsheet</a>
+			<iframe width=100% height=700 src="{{ link }}&rm=minimal"></iframe>
+		{% else %}
+			<h2>✈ Check back later ✈</h2>
+		{% endif %}
+	</div>
+</div>
+
+{% endblock %}
--- a/donut/modules/flights/templates/settings.html
+++ b/donut/modules/flights/templates/settings.html
@@ -1,0 +1,28 @@
+{% extends "layout.html" %}
+{% block page %}
+<div class="container theme-showcase" role="main">
+<div class="jumbotron">
+	<div class="col-sm-1">
+		    <a href="{{url_for('flights.flights')}}" class = "btn btn-primary back"> &lt;</a>
+	</div>
+	<div class="col-md-4">
+		<h2>Airport Carpool Settings</h2>
+		<form method="POST" action="{{ url_for('flights.update') }}">
+			<div class='form-group'>
+                		<label for="visible">Sheets are visible</label>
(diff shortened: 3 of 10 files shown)


sql/flights.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE IF EXISTS flights;
+
+CREATE TABLE flights (
+    id		INT DEFAULT 1,
```
calebsander: Is this table only being used to store a single row? If so, I would get rid of the `id` column and change the `INSERT INTO ...` SQL into a `DELETE` followed by an `INSERT INTO`.

sql/flights.sql
```diff
@@ -0,0 +1,8 @@
...
+    link	VARCHAR(250),
+    visible     BOOLEAN DEFAULT FALSE,
```
calebsander: What's the purpose of `visible`? It seems like we could just leave the table empty if we don't want to show a link.

sql/flights.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE IF EXISTS flights;
+
+CREATE TABLE flights (
+    id		INT DEFAULT 1,
+    link	VARCHAR(250),
```
calebsander: This should probably be `NOT NULL`; what would it mean to have a `NULL` link? Also links can be quite a bit longer than this, so it would probably be better to store this as a `TEXT` field instead.

donut/modules/flights/templates/flights.html
```diff
@@ -0,0 +1,19 @@
...
+		<h1>Airport Carpools</h1>
+		{% if admin %}                        
+			<a class="btn btn-primary" href="{{ url_for('flights.settings') }}" role="button">Settings</a>
+		{% endif %}
+		{% if not visible %}
+			<h2>&#9992; Check back later &#9992;</h2>
+		{% else %}
+			<a class="btn btn-primary" href="{{ link }}" target="_blank">Go to Spreadsheet</a>
+			<iframe width=100% height=700 src="{{ link }}&rm=minimal"></iframe>
+		{% endif %}
```
calebsander: Would be a little clearer to switch the `if` and `else` blocks here

donut/modules/flights/templates/flights.html
```diff
@@ -0,0 +1,19 @@
...
+{% block page %}
+
+<div class="container theme-showcase" role="main">
+	<div class="jumbotron">
+		<h1>Airport Carpools</h1>
+		{% if admin %}                        
+			<a class="btn btn-primary" href="{{ url_for('flights.settings') }}" role="button">Settings</a>
+		{% endif %}
+		{% if not visible %}
+			<h2>&#9992; Check back later &#9992;</h2>
```
calebsander: I think you should be able to put Unicode characters directly in the template rather than using HTML escapes

donut/modules/flights/templates/settings.html
```diff
@@ -0,0 +1,27 @@
...
+	<div class="col-md-4">
+		<h2>Airport Carpool Settings</h2>
+		<form method="POST" action="{{ url_for('flights.update') }}">
+                	<label for="visible">Sheets are visible</label>
+			<input type="checkbox" id="visible" name="visible" {% if visible %} checked {% endif %}>
+			<br>
+			<label for="link">Google Sheet link:</label>
+			<ol>
+				<li> File &rarr; Publish to the Web &rarr; Embed (ignore this link!)</li>
+				<li> File &rarr; Share &rarr; Copy link </li>
```
calebsander: Ditto on using Unicode characters here

donut/modules/flights/templates/settings.html
```diff
@@ -0,0 +1,27 @@
...
+<div class="jumbotron">
+	<div class="col-sm-1">
+		    <a href="{{url_for('flights.flights')}}" class = "btn btn-primary back"> &lt;</a>
+	</div>
+	<div class="col-md-4">
+		<h2>Airport Carpool Settings</h2>
+		<form method="POST" action="{{ url_for('flights.update') }}">
+                	<label for="visible">Sheets are visible</label>
+			<input type="checkbox" id="visible" name="visible" {% if visible %} checked {% endif %}>
+			<br>
```
calebsander: I think you can wrap inputs in `<div class='form-group'>` so they will be spaced out appropriately

calebsander (changes requested):

sql/flights.sql
rlin0: I was thinking if they don't want to change the link but just want to hide it during non break times so that people's info isn't easily visible, and after breaks the sheets are old/unnecessary info that doesn't need to be displayed. And I thought it'd be better to have a 'check again later' page than a blank table. idk

calebsander: Ah I meant the SQL table, not the spreadsheet. I think it would be fine to just delete the link when it's out of season.

rlin0: I was thinking like during season the sheet is visible and off season the sheet is hidden and for the next season the secretary can just make the sheet visible again and delete everything on the sheet. Like they don't have to input the link again if they want to reuse the same sheet for its headers or whatnot

calebsander: Sure, I don't feel super strongly about it, but it doesn't seem like it would be too problematic to just have them look up the link of the spreadsheet again if they have to reset the sheet anyways.

donut/modules/flights/routes.py
```diff
@@ -0,0 +1,35 @@
...
+    settings = helpers.get_settings()
+    return flask.render_template(
+        'flights.html',
+        visible=settings['visible'],
+        link=settings['link'],
+        admin=helpers.is_admin())
+
+
+@blueprint.route('/flights/update', methods=['POST'])
+def update():
```
calebsander, line 20: This route should verify `helpers.is_admin()` too

calebsander (dismissed):
Looks good to me, just make sure to check permissions on the `update` route

calebsander (approved):

PR #200, opened 2020-08-02
Allow admin to send and control all groups

rlin0:
### Summary
- Remove all rows in `position_relations` where `pos_id_from=pos_id` for devteam member 
- Check `is_admin` instead since devteam member should have access to any position
- Allow devteam member to post as devteam member to any group

### Test Plan
- Tested on localhost

### DIFF
--- a/donut/modules/newsgroups/routes.py
+++ b/donut/modules/newsgroups/routes.py
@@ -28,11 +28,12 @@ def post(group_id=None):
     user_id = auth_utils.get_user_id(flask.session['username'])
     if not group_id:
         group_id = flask.request.form.get('group')
+    if auth_utils.is_admin():
+        groups = helpers.get_newsgroups()
+    else:
+        groups = helpers.get_my_newsgroups(user_id, True)
     return flask.render_template(
-        'post.html',
-        groups=helpers.get_my_newsgroups(user_id, True),
-        group_selected=group_id,
-        page='post')
+        'post.html', groups=groups, group_selected=group_id, page='post')
 
 
 @blueprint.route('/newsgroups/postmessage', methods=['POST'])
@@ -45,7 +46,7 @@ def post_message():
     for field in fields:
         data[field] = flask.request.form.get(field)
     group_id = data['group']
-    actions = helpers.get_user_actions(user_id, group_id)
+    actions = get_user_actions(user_id, group_id)
     if not actions['send']:
         return flask.abort(403)
     data['group_name'] = groups.get_group_data(group_id,
@@ -80,7 +81,7 @@ def view_group(group_id):
     if 'username' not in flask.session:
         return flask.abort(403)
     user_id = auth_utils.get_user_id(flask.session['username'])
-    actions = helpers.get_user_actions(user_id, group_id)
+    actions = get_user_actions(user_id, group_id)
     fields = [
         'group_id', 'group_name', 'group_desc', 'anyone_can_send',
         'members_can_send', 'visible', 'admin_control_members'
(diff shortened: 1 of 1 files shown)


donut/modules/newsgroups/routes.py
```diff
@@ -178,4 +179,16 @@ def posting_positions(group_id):
     if 'username' not in flask.session:
         return flask.abort(403)
     user_id = auth_utils.get_user_id(flask.session['username'])
-    return flask.jsonify(helpers.get_posting_positions(group_id, user_id))
+    if auth_utils.is_admin():
+        pos = [{
+            'pos_id': groups.get_position_id('Devteam', 'Devteam Member'),
+            'pos_name': 'Devteam Member'
+        }]
+    pos.extend(helpers.get_posting_positions(group_id, user_id))
```
calebsander, line 187: It looks like `pos` will be undefined here if the user is not an admin. Make sure to test everything as a non-admin as well.

donut/modules/newsgroups/routes.py
```diff
@@ -178,4 +179,16 @@ def posting_positions(group_id):
     if 'username' not in flask.session:
         return flask.abort(403)
     user_id = auth_utils.get_user_id(flask.session['username'])
-    return flask.jsonify(helpers.get_posting_positions(group_id, user_id))
+    if auth_utils.is_admin():
+        pos = [{
+            'pos_id': groups.get_position_id('Devteam', 'Devteam Member'),
```
calebsander, line 184: Pretty sure the position name is just `Member`, not `Devteam Member`, so this is probably returning `None`. Also, what would happen if you submit `None` as the posting position?

donut/modules/newsgroups/routes.py
```diff
@@ -178,4 +179,16 @@ def posting_positions(group_id):
...
+            'pos_name': 'Devteam Member'
+        }]
+    pos.extend(helpers.get_posting_positions(group_id, user_id))
+    return flask.jsonify(pos)
+
+
+def get_user_actions(user_id):
+    if auth_utils.is_admin():
+        return {'send': True, 'control': True}
```
calebsander, line 193: This denies the `receive` permission if the user is an admin. Will that prevent admins from viewing messages sent to their newsgroups?

donut/modules/newsgroups/routes.py
```diff
@@ -178,4 +179,16 @@ def posting_positions(group_id):
...
+    if auth_utils.is_admin():
+        pos = [{
+            'pos_id': groups.get_position_id('Devteam', 'Devteam Member'),
+            'pos_name': 'Devteam Member'
+        }]
+    pos.extend(helpers.get_posting_positions(group_id, user_id))
+    return flask.jsonify(pos)
+
+
+def get_user_actions(user_id):
```
calebsander, line 191: I'm not a big fan of having two identically named functions. Is there a reason not to include the admin check in `helpers.get_user_actions()`? If the original behavior of `helpers.get_user_actions()` needs to be preserved, let's rename this function and move it to `helpers`.

calebsander (changes requested):

donut/modules/newsgroups/routes.py
rlin0: Because the unit tests for helper functions can't call auth_utils.is_admin()

donut/modules/newsgroups/routes.py
rlin0: Anyone who is a member can view the group. If `receive` is true, the page will display that the user is subscribed which can be false for admins

calebsander: Oh hmm it seems like that might be a bug that positions without `receive` can still see newsgroup messages, but that doesn't need to be addressed now

PR #202, opened 2020-09-08
Remove members_can_send field in groups

rlin0:
### Summary
- `members_can_send` is not used anywhere
- Whether a user can send to a group is controlled by `send` field in `positions` table
- will remove from MariaDB after

### Test Plan
- Unit tests


calebsander (approved):
Makes sense to me. Interesting to note that the `create_group()` helper function is not actually used anywhere :)

PR #203, opened 2020-09-09
Make newsgroups less ugly

rlin0:
### Summary
- Changes to UI to make it easier to read and less cluttered
Screen Shot 2020-09-08 at 5 29 16 PM

### Test Plan
- Manually tested out UI

### DIFF
--- a/donut/modules/newsgroups/helpers.py
+++ b/donut/modules/newsgroups/helpers.py
@@ -203,7 +203,16 @@ def get_owners(group_id):
     """
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, group_id)
-        return cursor.fetchall()
+        res = cursor.fetchall()
+        owners = {}
+        for row in res:
+            owners.setdefault(row['pos_name'], []).append({
+                'user_id':
+                row['user_id'],
+                'full_name':
+                row['full_name']
+            })
+        return owners
 
 
 def get_posting_positions(group_id, user_id):
--- a/donut/modules/newsgroups/templates/group.html
+++ b/donut/modules/newsgroups/templates/group.html
@@ -35,7 +35,7 @@
 					{% if not member %}
 						<li>
 							<form method="POST" id="apply" action="{{ url_for('newsgroups.apply_subscription', group_id=group['group_id']) }}">
-								<a href="#" onclick="this.parentNode.submit()">Apply for Subscription'</a>
+								<a href="#" onclick="this.parentNode.submit()">Apply for Subscription</a>
 							</form>
 						</li>
 					{% endif %}
--- a/donut/modules/newsgroups/templates/group.html
+++ b/donut/modules/newsgroups/templates/group.html
@@ -1,83 +1,121 @@
 {% extends "layout.html" %}
 {% block page %}
-{% include "navbar.html" %}
-<div class="container theme-showcase" role="main">
-  <div class="newsgroup-frame">
-
-{% if not group['visible'] and not member %}
-  <h2> Must be a member to view </h2>
-{% else %}
-  <h2> {{ group['group_name'] }} </h2>
-  {% if group['group_desc'] %}
-    <h3> {{ group['group_desc'] }} </h3>
-  {% endif %}
-  {% if not member %}
-    <form method="POST" id="apply" action="{{ url_for('newsgroups.apply_subscription', group_id=group['group_id']) }}">
-      <div class="form-group">
-	<input class="btn btn-primary" id="apply-btn" type="submit" value="Apply for Subscription"/>
-     </form>
-  {% endif %}
-  <ul style="list-style-type:disc;">
-  {% if group['anyone_can_send'] %}
-    <li> Anyone can post to this group. </li>
-  {% else %}
-    <li> Posting is restricted to authorized members. </li>
-  {% endif %}
-  {% if group['admin_control_members'] %}
-    <li> Administrators control the membership of this group. </li>
-  {% endif %}
-  {% if group['visible'] %}
-    <li> Anyone can learn that this group exists. </li>
-  {% endif %}
-  <li> You are {% if not actions['receive'] %} not {% endif %} subscribed. </li>
-  {% if actions["control"] %}
-    <li> You have admin control. </li>
-    <h4> Applications </h4>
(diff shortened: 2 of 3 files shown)


donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
-          </tr>
-        {% endfor %}
-    </table>
-    <a href={{ url_for('newsgroups.all_posts', group_id=group['group_id']) }}>View All</a>
-  {% else %}
-    <h4> Must be a member to view messages </h4>
-  {% endif %}
-{% endif %}
-</div>
+		{% if not group['visible'] and not member %}
```
calebsander: It would be great to keep the `{% if %}`, `{% else %}`, and `{% endif %}` blocks aligned (e.g. this `{% if %}` ends at the `{% endif %}` all the way at the end), and the HTML inside the blocks indented. Might be more readable to put the code for the `{% else %}` block (i.e. `group['visible'] or member`) in a separate template file. And I personally find `{% if group['visible'] or member %} ... {% else %} ... {% endif %}` more readable, but your call

donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
+		<h2> Must be a member to view </h2>
+		{% else %}
+		<h2> {{ group['group_name'] }}
+			{% if actions['send'] or group['anyone_can_send'] %}
+			<a href={{ url_for('newsgroups.post', group_id=group['group_id']) }}><span class="glyphicon glyphicon-send"></span></a>
```
calebsander: Hmm I worry it's not super clear that this icon is a link. Maybe style it as a button (i.e. add `class="btn btn-default"` on the anchor tag)?

donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
+		</h2>
+		{% if group['group_desc'] %}
+		<h3> {{ group['group_desc'] }} </h3>
+		{% endif %}
+		<ul style="list-style-type:disc;">
+			{% if group['anyone_can_send'] %}
+			<li> Anyone can post to this group. </li>
+			{% else %}
+			<li> Posting is restricted to authorized members. </li>
+			{% endif %}
```
calebsander: Nit, but could move the `<li>` tags outside the `{% if %}` and `{% else %}` clauses

donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
+						{% endfor %}
+					</table>
+					{% else %}
+					You do not have access.
+					{% endif %}
+				</div>
+			</div>
+			<div class="panel panel-default">
+				<div class="panel-heading">
+					<h4 class="panel-title">
```
calebsander: Missing `</h4>`

donut/modules/newsgroups/helpers.py
```diff
@@ -203,7 +204,14 @@ def get_owners(group_id):
     """
     with flask.g.pymysql_db.cursor() as cursor:
         cursor.execute(query, group_id)
-        return cursor.fetchall()
+        res = cursor.fetchall()
+        owners = defaultdict(list)
```
calebsander: I would try to avoid unusual Python classes like this. Could just use [`dict.setdefault(row['pos_name'], []).append(...)`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault).

donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
+			{% endif %}
+			<li> You are {% if not actions['receive'] %} not {% endif %} subscribed. </li>
+			{% if actions["control"] %}
+			<li> You have admin control. </li>
+			{% endif %}
+			{% if not member %}
+			<li>
+				<form method="POST" id="apply" action="{{ url_for('newsgroups.apply_subscription', group_id=group['group_id']) }}">
+					<div class="form-group">
+						<a href="#" onclick="$(this).closest('form').submit();">Apply for Subscription</>
```
calebsander: Should be `</a>`

calebsander: Also, can use `<button type='submit'>` instead to avoid the `onclick` handler

donut/modules/newsgroups/templates/group.html
```diff
@@ -2,82 +2,118 @@
...
+			<div class="panel panel-default">
+				<div class="panel-heading">
+					<h4 class="panel-title">
+						<a data-toggle="collapse" href="#owners">Group Owners</a> 
+				</div>
+				<div id="owners" class="panel-body collapse">
+					{% for pos_name, pos_people in owners.items() %}
+					<h5> {{ pos_name }}: </h5>
+					{% for person in pos_people %} 
+					<a href={{ url_for('directory_search.view_user', user_id=person.user_id) }}>{{ person.full_name }}</a><br>
```
calebsander: Prefer wrapping `<a>` in `<div>` instead of adding manual line breaks

calebsander (changes requested):

donut/modules/newsgroups/templates/group.html
rlin0: I did that because I wanted it to be text because the button doesn't fit the aesthetic lol.

donut/modules/newsgroups/templates/group.html
rlin0: I feel like the button ruins the aesthetics lol 😬

donut/modules/newsgroups/templates/group.html
calebsander: `class="btn btn-link"` makes a button render like a link

donut/modules/newsgroups/templates/group.html
calebsander: Okay, it's probably fine as long as the people who send stuff know about it...

donut/modules/newsgroups/templates/group.html
rlin0: still looks like a button. I could add css to make the button look like a link or just keep `<a href="#" onclick="this.parentNode.submit()"`

calebsander (approved):

PR #204, opened 2020-09-11
Allow admin access to any group

rlin0:
### Summary
- Remove all rows in `position_relations` where `pos_id_from=pos_id` for devteam member
- Check `is_admin` instead since devteam member should have access to any position
- Allow devteam member to post as devteam member to any group

### Test Plan
- Unit tests and tested with fake `test_newsgroup`

### DIFF
--- a/donut/auth_utils.py
+++ b/donut/auth_utils.py
@@ -258,6 +258,15 @@ def get_user_id(username):
     return user and user['user_id']
 
 
+def get_username(user_id):
+    """Takes a user_id and returns the user's username."""
+    query = 'SELECT username FROM users WHERE user_id = %s'
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, [user_id])
+        user = cursor.fetchone()
+    return user and user['username']
+
+
 def update_last_login(username):
     """Updates the last login time for the user."""
     query = """
@@ -359,9 +368,31 @@ def is_caltech_user():
         re.search(ip_match, flask.request.remote_addr) is not None
 
 
-def is_admin():
+def is_admin(user_id=None):
     """
     Returns whether the current user is an admin
     """
-    username = flask.session.get('username')
+    if user_id:
+        username = get_username(user_id)
+    else:
+        username = flask.session.get('username')
     return username and check_permission(username, None)
+
+
+def get_admin_pos(user_id):
+    """
+    Returns position id and name of the admin position.
+    """
+    if not is_admin(user_id):
+        return []
+    query = """
+        SELECT pos_id, CONCAT(group_name, ' ', pos_name) AS pos_name 
+        FROM position_permissions 
+        NATURAL JOIN positions
+        NATURAL JOIN groups
+        NATURAL JOIN current_direct_position_holders
+        WHERE user_id=%s AND permission_id=%s
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (user_id, Permissions.ADMIN.value))
+        return cursor.fetchall()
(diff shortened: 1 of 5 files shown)


donut/auth_utils.py
```diff
@@ -359,9 +368,31 @@ def is_caltech_user():
         re.search(ip_match, flask.request.remote_addr) is not None
 
 
-def is_admin():
+def is_admin(user_id=None):
```
calebsander, line 371: If `user_id` is passed in, we should be able to pass it directly to `check_permission()` (and `get_permissions()` as well). Currently, this code looks up the username from the user ID and then looks up the user ID from the username. Ideally we should find all calls to `check_permission()` where the user ID could be passed in directly (to avoid a DB lookup).

donut/auth_utils.py
```diff
@@ -359,9 +368,31 @@ def is_caltech_user():
...
+        username = flask.session.get('username')
     return username and check_permission(username, None)
+
+
+def get_admin_pos(user_id):
+    """
+    Returns position id and name of the admin position.
+    """
+    if not is_admin(user_id):
+        return []
```
calebsander, line 387: Is this check needed? I don't think the DB query below will return any rows if the user is not an admin. DB queries are generally slow, so it would be good to avoid them if possible.

donut/modules/newsgroups/helpers.py
```diff
@@ -65,6 +68,9 @@ def get_user_actions(user_id, group_id):
         cursor.execute(query, (user_id, group_id))
         res = cursor.fetchall()
     actions = {'send': False, 'control': False, 'receive': False}
+    if auth_utils.is_admin(user_id):
+        actions['send'] = True
+        actions['control'] = True
```
calebsander, line 73: Maybe a little clearer to write:
```py
is_admin = auth_utils.is_admin(user_id)
actions = {'send': is_admin, 'control': is_admin, 'receive': False}
```

donut/modules/newsgroups/routes.py
```diff
@@ -90,7 +90,8 @@ def view_group(group_id):
     if actions['control']:
         applications = helpers.get_applications(group_id)
     messages = None
-    member = groups.is_user_in_group(user_id, group_id)
+    member = auth_utils.is_admin() or groups.is_user_in_group(
```
calebsander, line 93: If you change the behavior of `is_admin()` to avoid looking up the username from the user ID, it would be more efficient to pass `user_id` here

donut/modules/newsgroups/helpers.py
```diff
@@ -209,6 +215,7 @@ def get_owners(group_id):
 def get_posting_positions(group_id, user_id):
     """Get positions user can send as in a group."""
 
+    pos = auth_utils.get_admin_pos(user_id)
```
calebsander, line 218: Hmm, to reduce code duplication maybe it makes sense for all these queries to return `pos_id, group_name, pos_name` and then to concatenate the group and position names in Python.

calebsander (changes requested):
Hmm, I realize I said this was a good idea earlier, but I'm not so sure now that I see all the places where the code needs to be changed. And there are at least a few other places where Devteam members will look different from other `ug` admins: our directory pages and the campus positions page will no longer show these indirect positions. Honestly, I think `is_admin()` is a less elegant way to give Devteam posting/admin privileges than using the position relations mechanism we have created for this purpose. Just my 2 cents, let me know what you think.

The code changes look good to me except for some concerns about unnecessary DB queries.

rlin0:
I guess it makes sense to use position_relations, so for admins they would just have position_relations to every position? And if new positions are added, we have to remember to add the position_relation for admins?

calebsander:
Yeah in an ideal world I think we would define a permission for each thing admins can do and give the devteam member position those permissions. (I think checking `is_admin()` is a pretty hacky way to tell whether a user should be able to add news, for example.) But it would be a lot of work, and I don't see a lot of situations where that extensibility would actually be needed in the future.

rlin0:
Use position_relations instead of adding a special case for admins

PR #209, opened 2020-10-03
restrict posting feedback to caltech users

rlin0:
### Summary
- Restrict submitting feedback to logged in users or users on Caltech VPN

### Test Plan
- localhost

### DIFF
--- a/donut/modules/feedback/routes.py
+++ b/donut/modules/feedback/routes.py
@@ -15,6 +15,8 @@ permissions = {
 
 @blueprint.route('/feedback/<group>')
 def feedback(group):
+    if not auth_utils.is_caltech_user():
+        return auth_utils.login_redirect()
     if group not in Groups:
         return flask.abort(404)
     summary = False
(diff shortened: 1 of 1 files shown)


donut/modules/feedback/routes.py
```diff
@@ -15,7 +15,7 @@
 
 @blueprint.route('/feedback/<group>')
 def feedback(group):
-    if group not in Groups:
+    if group not in Groups or not auth_utils.is_caltech_user():
```
calebsander: I think `not (group in Groups and auth_utils.is_caltech_user())` is a bit clearer, but your call

donut/modules/feedback/routes.py
```diff
@@ -15,7 +15,7 @@
...
         return flask.abort(404)
```
calebsander, line 21: This should show a "you need to login" message, not a 404 error. See how the directory search pages do it: https://github.com/ASCIT/donut/blob/3181076b26c91db50177924794b1309af0572f04/donut/modules/directory_search/routes.py#L15-L16

calebsander (changes requested):
Please suggest logging in instead of just showing a 404 page

calebsander (approved):

PR #210, opened 2020-10-12
Add feature that allows users to save notes for their courses

rlin0:
### Summary
- Some courses use Zoom, Canvas, and their own website so it's annoying to keep track of all these links
- Allow users to enter and save notes for each class in their scheduler
- links are converted into clickable href elements

### Test Plan
- Tested UI

### DIFF
--- a/donut/modules/courses/helpers.py
+++ b/donut/modules/courses/helpers.py
@@ -303,3 +303,35 @@ def get_user_scheduler_sections(username, year, term):
 
 is_duplicate_error = lambda e: \
     isinstance(e, IntegrityError) and e.args[0] == ER.DUP_ENTRY
+
+
+def get_notes(username, course):
+    user_id = get_user_id(username)
+    query = "SELECT notes FROM scheduler_notes WHERE user_id=%s AND course_id=%s"
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (user_id, course))
+        res = cursor.fetchone()
+        if res:
+            return res['notes']
+        else:
+            return None
+
+
+def edit_notes(username, course, notes):
+    if len(notes) == 0:
+        delete_notes(username, course)
+        return
+    user_id = get_user_id(username)
+    query = """
+    INSERT INTO scheduler_notes (user_id, course_id, notes) VALUES (%s, %s, %s)
+    ON DUPLICATE KEY UPDATE notes = VALUES(notes)
+    """
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (user_id, course, notes))
+
+
+def delete_notes(username, course):
+    user_id = get_user_id(username)
+    query = "DELETE FROM scheduler_notes WHERE user_id=%s AND course_id=%s"
+    with flask.g.pymysql_db.cursor() as cursor:
+        cursor.execute(query, (user_id, course))
--- a/donut/modules/courses/routes.py
+++ b/donut/modules/courses/routes.py
@@ -197,5 +197,6 @@ def edit_notes():
 @blueprint.route('/1/scheduler/notes/<int:course>')
 def get_notes(course):
     username = flask.session.get('username')
-    if not username: return flask.jsonify(())
+    if not username: return flask.jsonify('')
+
     return flask.jsonify(helpers.get_notes(username, course) or '')
--- a/donut/modules/courses/routes.py
+++ b/donut/modules/courses/routes.py
@@ -178,3 +178,24 @@ def scheduler_mine(year, term):
 
     return flask.jsonify(
         helpers.get_user_scheduler_sections(username, year, term))
+
+
+@blueprint.route('/1/scheduler/edit_notes', methods=['POST'])
+def edit_notes():
+    username = flask.session.get('username')
+    if not username:
+        return flask.jsonify({
+            'success': False,
+            'message': 'Must be logged in to save'
+        })
+    helpers.edit_notes(username,
+                       flask.request.form.get('courseId'),
+                       flask.request.form.get('notes'))
+    return flask.jsonify({'success': True})
+
+
+@blueprint.route('/1/scheduler/notes/<int:course>')
+def get_notes(course):
+    username = flask.session.get('username')
+    if not username: return flask.jsonify(())
+    return flask.jsonify(helpers.get_notes(username, course) or '')
--- a/donut/modules/courses/templates/scheduler.html
+++ b/donut/modules/courses/templates/scheduler.html
@@ -1,5 +1,6 @@
 {% extends 'layout.html' %}
 {% block page %}
+  {% include "notes_modal.html" %}
   <div class='col-md-10 col-md-offset-1' style='float: none'>
     <h3 id='warning'></h3>
     <h1>Course Scheduler</h1>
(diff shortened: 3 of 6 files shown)


donut/modules/courses/helpers.py
```diff
@@ -303,3 +303,25 @@ def get_user_scheduler_sections(username, year, term):
...
+            return res['notes']
+        else:
+            return None
+
+
+def edit_notes(username, course, notes):
+    user_id = get_user_id(username)
+    query = """
+    INSERT INTO scheduler_notes (user_id, course_id, notes) VALUES (%s, %s, %s)
+    ON DUPLICATE KEY UPDATE notes=%s
```
calebsander: You can avoid passing `notes` twice by doing `ON DUPLICATE KEY UPDATE notes = VALUES(notes)`

donut/modules/courses/routes.py
```diff
@@ -178,3 +178,24 @@ def scheduler_mine(year, term):
...
+def edit_notes():
+    username = flask.session.get('username')
+    if not username:
+        return flask.jsonify({
+            'success': False,
+            'message': 'Must be logged in to save'
+        })
+    helpers.edit_notes(username,
+                       flask.request.form.get('courseId'),
+                       flask.request.form.get('notes'))
```
calebsander, line 193: Is there a way to delete notes? Maybe they should be deleted if set to the empty string?

donut/modules/courses/templates/scheduler.html
```diff
@@ -258,7 +271,19 @@ <h1>Course Scheduler</h1>
...
+              $('<button>')
+                .attr({
+                  type: 'button',
+                  course: course.id,
+                  section: sectionNumber,
+		  class: 'btn edit',
+		  "data-toggle": 'modal',
+		  "data-target": '#modal',
+		  "data-course-id": course.id,
+		  "data-course-num": course.number
```
calebsander: Nit: mixing tabs and spaces here (and on the lines changed above)

donut/modules/courses/templates/scheduler.html
```diff
@@ -461,7 +486,58 @@ <h1>Course Scheduler</h1>
     function displayUnits() {
       units.text('Units: ' + String(coursesUnits))
     }
-
+  $('#modal').on('show.bs.modal', function(e) {
+    var courseNum = $(e.relatedTarget).data("courseNum");
+    var courseId = $(e.relatedTarget).data("courseId");
+    displayNotes(courseNum, courseId); 
+  });
+  $('#editButton').click(function(e) {
```
calebsander: Can omit `e` if you're not using it

donut/modules/courses/templates/scheduler.html
```diff
@@ -461,7 +486,58 @@ <h1>Course Scheduler</h1>
...
+    $('#editButton').hide();
+  });
+  $('#postNotes').submit(function(e) {
+    e.preventDefault();
+    $.ajax({
+      url: '/1/scheduler/edit_notes',
+      type: 'POST',
+      data:$('#postNotes').serialize(),
+      success:function(){
+        displayNotes($('#courseNum').val(), $('#courseId').val());
```
calebsander: If this returns an error message, you should display it

donut/modules/courses/templates/scheduler.html
```diff
@@ -461,7 +486,58 @@ <h1>Course Scheduler</h1>
...
+  }
+  function URLify(string) {
+    const urls = string.match(/(((ftp|https?):\/\/)[\-\w@:%_\+.~#?,&\/\/=]+)/g);
+    if (urls) {
+      urls.forEach(function (url) {
+        string = string.replace(url, '<a target="_blank" href="' + url + '">' + url + "</a>");
+      });
+    }
+    return string.replace("(", "<br/>(");
+  }
```
calebsander: Use `format-text.js` instead? Injecting unsanitized strings as HTML is not a great practice, even if users can only do it to themselves

sql/courses.sql
```diff
@@ -80,3 +80,13 @@ CREATE TABLE scheduler_sections (
...
         ON DELETE CASCADE
 );
+
+CREATE TABLE scheduler_notes (
+    user_id         INT      NOT NULL,
+    course_id       INT      NOT NULL,
+    notes	    TEXT     DEFAULT NULL,
+    PRIMARY KEY (user_id, course_id),
+    FOREIGN KEY (user_id, course_id) 
+	REFERENCES scheduler_sections(user_id, course_id)
```
calebsander: `(user_id, course_id)` is not a unique key for `scheduler_sections` since a user can add multiple sections of the same course

sql/courses.sql
```diff
@@ -80,3 +80,13 @@ CREATE TABLE scheduler_sections (
         REFERENCES sections(course_id, section_number)
         ON DELETE CASCADE
 );
+
+CREATE TABLE scheduler_notes (
+    user_id         INT      NOT NULL,
+    course_id       INT      NOT NULL,
+    notes	    TEXT     DEFAULT NULL,
```
calebsander: What would `notes = NULL` mean? I think this column should be `NOT NULL`.

donut/modules/courses/templates/scheduler.html
```diff
@@ -461,7 +486,58 @@ <h1>Course Scheduler</h1>
...
+  });
+  $('#editButton').click(function(e) {
+    $('#postNotes').show();
+    $('#viewNotes').hide();
+    $('#editButton').hide();
+  });
+  $('#postNotes').submit(function(e) {
+    e.preventDefault();
+    $.ajax({
+      url: '/1/scheduler/edit_notes',
```
calebsander: Use `url_for()`?

donut/modules/courses/templates/scheduler.html
```diff
@@ -461,7 +486,58 @@ <h1>Course Scheduler</h1>
...
+      data:$('#postNotes').serialize(),
+      success:function(){
+        displayNotes($('#courseNum').val(), $('#courseId').val());
+      },
+      error: function() {
+        displayWarning('Failed to save notes')
+      }
+    });
+  });
+  function displayNotes(courseNum, courseId) {
```
calebsander: Just clarifying—are notes hidden until you click the section on the scheduler?

calebsander (changes requested):

donut/modules/courses/helpers.py
rlin0: Doesn't VALUES(notes) = the old notes value?

donut/modules/courses/templates/scheduler.html
rlin0: Yes they're displayed in a modal

donut/modules/courses/helpers.py
calebsander: No, that's just `notes`. `VALUES(notes)` is the new value of `notes` you are trying to insert. We already do this in the directory update script: https://github.com/ASCIT/donut/blob/master/donut/modules/directory_search/utils/import_registrar.py#L88-L91

donut/modules/courses/templates/scheduler.html
rlin0: I was thinking to keep it consistent with the old code which does not use url_for

donut/modules/courses/templates/scheduler.html
rlin0: Like this? https://github.com/ASCIT/donut/blob/1319a70e3f221b70e57235200c22d003a183f13c/donut/modules/uploads/static/script_uploading.js#L28

calebsander: No, it will still be in the `success` callback, but just check `data.success` and display `data.message` if it is false. I believe I do this in some of the other AJAX calls.

calebsander (approved):

PR #217, opened 2020-11-28
Add directory search by timezone 

rlin0:
### Summary
- Users can add their timezone to their profile
- Directory search by timezone range
- Feature request from ASCIT for students looking to collaborate with people in the same time zone

### Test Plan
- Tested UI


calebsander (approved):
I made a few changes to allow timezone offsets in minutes and sort the timezone list by offset. Also did some refactoring to reduce code duplication. Looks good to me otherwise.

Make sure to `ALTER TABLE members ...` before pulling the changes to prod

PR #237, opened 2021-11-22
Limit uploads to 10 MB (instead of 1 GB)

calebsander:
Currently, the database crashes when we try to feed it a directory image much larger than 10 MB. This leads to an "500 Internal Server Error" message, which is very unhelpful. It also makes the website unusable for other users until the database can be restarted, which is really bad.

So use a lower upload size limit and show a more helpful 413 error page if the upload is too large.

### Test Plan

Tried uploading a 15 MB directory image and verified that it immediately returned a 413 error page saying the limit was 10 MB. Smaller images work normally. Verified unit change algorithm also works if the size limit exceeds the largest unit (e.g. will show 10240 KB instead).

### DIFF
--- a/donut/__init__.py
+++ b/donut/__init__.py
@@ -151,6 +151,14 @@ def access_forbidden(error):
     return flask.render_template("403.html"), http.client.FORBIDDEN
 
 
+@app.errorhandler(http.client.REQUEST_ENTITY_TOO_LARGE)
+def upload_too_large(error):
+    """ Handles a 413 Payload Too Large error. """
+    max_size = constants.MAX_CONTENT_LENGTH_STRING
+    return flask.render_template("413.html", max_size=max_size), \
+        http.client.REQUEST_ENTITY_TOO_LARGE
+
+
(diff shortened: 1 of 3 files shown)


donut/__init__.py
```diff
@@ -151,6 +151,23 @@ def access_forbidden(error):
...
+    next_max_size = constants.MAX_CONTENT_LENGTH
+    for unit in UNITS:
+        max_size = next_max_size
+        next_max_size /= 1024
+        if next_max_size < 1:
+            break
+    int_max_size = int(max_size)
+    if int_max_size == max_size:
+        max_size = int_max_size
+    return flask.render_template("413.html", max_size=f'{max_size} {unit}'), \
```
rlin0: Why not just `return flask.render_template("413.html", max_size='10 MB')` ? What's the point of the code above?

calebsander: I was trying to avoid having to specify the limit in two different places, but I agree this unit computation is a bit overkill. Made the human-readable version of the size a constant next to `MAX_CONTENT_LENGTH` in the constants file.

